PR #57, opened 2015-11-10
Kain: Scribble HW (Wednesday & Friday)

elisekain:
Completeness: 5/5
Comfort Level: 4/5

Deployed Site: https://kain-scribble.herokuapp.com/

seed file used to work...doesn't work right now since I played around with user authentication and adding/removing columns to create relationships.


jsm13:
:+1:  Awesome!

PR #58, opened 2015-11-11
Joe gz scribble solution par1

joe-gz:
comfort: 4
completeness: 5 (for Wednesday)

I feel pretty good with this so far. I have very minimal css right now, as I want to get this working first. Will try to add some in for Friday!


ebirving:
Great stuff, Joe! Super thorough. I love your dedication to practicing the whole workflow, from planning on down through CSS, on every assignment. Quick hint: you might want to throw a `, dependent: :destroy` into your Post model -- might save you from some delete-related headaches down the line.

PR #59, opened 2015-11-11
Alice Frumin Solution-Scribble-Wednesday

alifrumin:
Completeness: 4
Comfort: 3

I feel like I learned a lot


ebirving:
:tada: Looks great, Alice! A couple of hints as you tackle the next segment... First, I'd recommend getting in the habit of committing early and often before we hit Project week on Monday. Second, you might want to throw a `, dependent: :destroy` in your Post model to guard against delete-related headaches.

PR #60, opened 2015-11-11
Pz2 views controllers - wednesday

pzhang87:
completion: 4 (hopefully)
comfort: 3.5

feeling good about RoR but still not super-confident; made various progress in different areas and hopefully met objectives but a little too fried to go back and check (hence, the 4).  need to be better about doing things in sprints.


nolds9:
:+1:

PR #62, opened 2015-11-11
Cpgruber solution

cpgruber:
Comfort_level:3.5
Completeness:4.5

I think I tried to do too much for what was required by Wednesday. I added helpers in some places, tried to use `form_for`, which I didn't get working, and tried nesting comment routes in posts, which just turned into more of a headache than I need right now, so I time-boxed and put it down. I think I will start from scratch for Friday's submission and integrate helpers and nesting from the get-go instead of having to recursively add them.


jshawl:
:+1:

PR #67, opened 2015-11-11
Chris_Kibble_Wednesday

ckib16:
Still need to do views & controllers w/ strong params


RobertAKARobin:
Boo, `rails g`! That should only be used with `rails g migration`. Otherwise it generates a boatload of `.scss` and `.coffee` files that you don't need -- hence why your repo has 60 more files than everyone else's!

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def new
+    @post = Post.new
+  end
+
+  def create
+    @post = Post.create!(post_params)
+      #assign result to @post just so can pass @post into redirect_to below to get the :id
+    redirect_to post_path(@post)
```
RobertAKARobin, line 14: You can also just `redirect_to @post`.

app/views/posts/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>New Post Form</h1>
+
+<%= form_for @post do |f| %>
+<%= f.label :title %>
+<%= f.text_field :title %>
```
RobertAKARobin, line 5: Indent!

RobertAKARobin:
Looking good so far!

PR #69, opened 2015-11-11
Rachel Wednesday hw

ddayporter:
Comfort: 4
Completeness: 5

I think I've got the minimum requirements for what's due tomorrow, and started laying the groundwork for adding users. I'm still leaning on the Tunr crutch quite a lot, as a reference to remember what to do.  Hopefully it will come more naturally with a bit more practice..


mattscilipoti:
:+1:
No need to generate the devise views.  Unless you want to change them.

PR #72, opened 2015-11-11
Starr's Scribble - ALL

starrchen:
For Wednesday --

comfort: 4
completeness: 5

Went ahead and used helpers and nested routes but, as I've run waaaay over my timebox, I haven't made sure it's all working yet hahaha. (Fingers crossed it is!)

I'm not sure I fully understand _strong params_, still, and I am a bit shaky on how the `created_at` works, so I just left it to not be edit-able or new-able.... hoping that it will still work? (I don't want the user to be changing or manually writing that -- for now -- anyway.) 

Also, just to clarify, form helpers don't need authenticity tokens, right? How can we explicitly write one in?

To be honest, I am having a hard time getting motivated or pleasure out of coding this week, especially given that so much of what we've been doing this week in class is undoing/changing what we did last week with Tunr. I know I have to work on not letting myself get too attached to code, regardless of how long I've spent on it. But I do know that yesterday I was hesitant to get much of the homework done because I feared/knew we'd just be undoing it today anyway. Here's hoping there's no more of that for the rest of the week.

---

For Friday --

comfort: 3
completeness: 5

~~Couldn't get deployment to work, and obviously am way over timebox here...~~
JUST KIDDING I got it: http://starr-scribble.herokuapp.com

I want to seed some user data in there so that I can associated seeded posts w/ user (to display the email of the author) but I can't figure out how to do that with devise involved.

Getting the authentication and the deployment has been one of the more painful things I've worked on so far, and the deployment didn't work at all... I feel like I understand what I'm supposed to do but I am a loss when I run into an error because I don't really have the debugging skills yet.

_ALSO_ I don't know if I am supposed to delete the `tmp` folder and the development logs or something? It just seems like a lot of information from testing that doesn't need to be saved. I have been deleting `tmp` files because they add just so many lines when I'm looking at `git status` and I don't want to deal with them.... (see commit `95f73b3`)


amaseda:
First off, love the quantity and titles of your commits. Great habit to get into.

amaseda:
**Re: users in `seeds.rb`**

Good question. Came across this: http://stackoverflow.com/questions/12418584/seeding-users-with-devise-in-ruby-on-rails

Going to mess around with this myself and see what the best way to do it. You're right, it would be nice not to have to create users via the browser.

amaseda:
**Re: deployment**

**UPDATE:** Noticing now that you were able to deploy successfully! :+1:

Would have to see the issue to diagnose it. Happy to sit down together and give it a shot.

Deployment errors are perhaps the most frustrating errors you will encounter in WDI. I still struggle with it plenty.

amaseda:
**Re: tmp and logs**

You can include those files/directories in your `.gitignore` so that they are not pushed to Github. They won't appear when you run `git status` either.

amaseda:
**Re: form helpers and tokens**

Correct, you do not need to pass in an authenticity token when using form helpers. If you want to explicitly pass one in then I would write it out in pure HTML [like so](http://stackoverflow.com/questions/4085136/how-to-generate-authenticitytoken-on-rails).

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,47 @@
+class PostsController < ApplicationController
+
+  def index
+    @posts = Post.all
+  end
+
+  def new
+    authenticate_user!
```
amaseda, line 8: You can make your code DRYer by using `before_action` to run `authenticate_user!` (vs. writing it out for each controller action).

config/secrets.yml
```diff
@@ -0,0 +1,22 @@
+# Be sure to restart your server when you modify this file.
```
amaseda, line 1: You don't want to push `secrets.yml` to a public repo! Files with sensitive information (e.g., API keys) should be included in your `.gitignore`.

amaseda:
Great job with this. Hope my comments answered your questions. Let me know if you have any more.

:+1:

PR #73, opened 2015-11-11
dimitri gorgan scribble (Wednesday update)

dgorgan:
comfort: 4
completeness: 4

Feeling ok up to this point, although i still need to finish up my Comment migration and to fill out Its necessary actions (all 7 of which are already defined, but empty) with the correct data. I did end up using link and form helpers for the homework however..

cue /giphy zoidberg you're bad and you should feel bad


ebirving:
I'm always in favor of a good Zoidberg gif, but there's no reason to feel bad here! Implementing helpers "early" is just efficient practice :stuck_out_tongue_winking_eye: What your have looks good, and you seem to know where you're going in re that empty Comments _Controller_. :+1:

PR #75, opened 2015-11-11
Samir - HW Scribble Part 1

samir-acle:
Comfort: 4
Completeness: 5

Questions: When defining instance variables in the comments-controller, is @post = @comment.post faster than @post = Post.find(params[:artist_id])?
which is better practice?
Is consistancy important?


RobertAKARobin:
If you already have `@comment` defined, I would do `@comment.post`. I _think_ it's probably faster -- but whether it is or it isn't, we're talking nanoseconds here. And yes, consistency is always a plus from a readability perspective.

scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.new
+  end
+
+  def update
+    @post = Post.find(params[:post_id])
+    @comment = Comment.find(params[:id])
+    @comment.update(comments_params)
+
+    redirect_to post_comment_path(@post, @comment)
```
RobertAKARobin, line 22: If every comment has a unique ID, then I would make a route for comments so that you can just `redirect_to @comment`. A route like `/artists/32/songs/3` is redundant, since both IDs are unique. Ignoring all the "favorites" stuff, this provides an example of what I'm talking about:

https://github.com/ga-dc/tunr_rails_deployment/blob/master/config/routes.rb

scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def update
+    @post = Post.find(params[:id])
+    @post.update(posts_params)
+    redirect_to @post
+  end
+
+  def destroy
+    @post = Post.find(params[:id])
```
RobertAKARobin, line 31: Could you write a `before_action` so you don't need to write `@post = Post.find(params[:id])` a zillion times?

scribble/app/views/comments/show.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2>Comment</h2>
+
+<p><%= @comment.text %></p>
+<p>Created At: <%= @comment.created_at %></p>
+<p>Updated At: <%= @comment.updated_at %></p>
```
RobertAKARobin, line 5: Be sure to take a look at `.strftime`, which will let you format the dates appropriately.

scribble/app/views/posts/edit.html.erb
```diff
@@ -0,0 +1,5 @@
+<%= render partial: "form" %>
+
+<%= form_for @post, method: "delete" do |f| %>
+  <%= f.submit "Delete this post" %>
+<% end %>
```
RobertAKARobin, line 5: You can also:

```
<%= link_to "Delete this post", @post, method: :delete %>
```

PR #77, opened 2015-11-11
SMC Wednesday Scribble

ghost:
Completeness 70% - ERD, Models, Schema and Seeds are set up in a ruby app. No authentication or any of third block of requirements.

Comfort: 2/5 unclear if my schema/seed files will work still just winging it..- having just email seems incomplete. but I thought a time stamp would be system generated by authentication tools, I can't keep track of what MVC tools are used for each of the 3-4 database mgmt methods we have learned.


scribble/db/seeds.rb
```diff
@@ -0,0 +1,15 @@
+# This fiTRUNCATE TABLE comments CASCADE;
+TRUNCATE TABLE comments CASCADE;
+
+ALTER SEQUENCE posts_id_seq RESTART WITH 1;
+ALTER SEQUENCE comments_id_seq RESTART WITH 1;
+
+
+INSERT INTO  (email) VALUES ('mrchensorganiccontrolroom@gmail.com' );
+
```
amaseda, line 9: Your `seeds.rb` file should only contain Ruby code. So instead of SQL, you're going to use ActiveRecord methods to populate/modify your database. Take a look at the examples that are included in the default commented out code in this file.

scribble/db/schema.rb
```diff
@@ -0,0 +1,14 @@
...
+  id SERIAL PRIMARY KEY,
+  email TEXT
+);
+
+CREATE TABLE comments (
+  id SERIAL PRIMARY KEY,
+  email TEXT,
+  post_id INTEGER
+
+);
```
amaseda, line 14: Echoing my comments left in the `seeds.rb` file. This shouldn't contain SQL. Instead, it will contain Ruby code that generates tables. You will never modify the `schema.rb` file yourself. You should always use and run migrations, which will define what your schema file looks like.

amaseda:
:+1: Will keep an eye out for your Friday pull request and leave more code comments on that.

ghost:
Hi,

Yes I will do this over the weekend. I will submit by 9AM- so that I can make my way to an MVP for next weeks project. I am slowly going through this stuff- way too slow- but I will get it all done...eventually.

PR #83, opened 2015-11-11
Tyler Crosse - Scribble progress

tylercrosse:
Completeness: 5/5
Comfort 4.5/5

Finally starting to feel ok about rails. It's been a journey :/

https://scribble-tyler.herokuapp.com/


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,49 @@
+class PostsController < ApplicationController
+  before_action :authenticate_user!, only: [:new, :create, :edit, :update, :destroy]
+
+  def index
+    @posts = Post.all
+  end
+
+  def new
+    @user = current_user
```
amaseda, line 9: Instead of creating `@user` multiple times, why not just reference `current_user` in the view? That'll save you some lines of code in your controller.

If you still want to keep `@user` though, you could use `before_action` to generate it automatically before each controller action (vs. writing it out each time).

app/models/user.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  after_create :add_identicon
+
+  def add_identicon
+    puts '*' * 50
+    puts self.email
+    @identicon = Identicon.data_url_for self.email, 48, [255, 255, 255]
+    puts @identicon
+    self.update(identicon: @identicon)
+  end
```
amaseda, line 20: This is cool! Going to look into identicon myself.

amaseda:
Great job here. Let me know if you have any additional questions about this assignment.

:+1:

PR #88, opened 2015-11-11
Jmoon solution

mooniker:
Completeness: 2/3
Comfort: 3/3

Didn't really tackle this in the prescribed order. Started with a nested structure so kinda got hung up with getting the comments to show on a post instead of their own separate page.

ERD is here: https://github.com/mooniker/scribble/blob/jmoon_solution/erd.jpg


andrewsunglaekim:
:+1:

PR #98, opened 2015-11-11
Andres solution

codedre:
completeness 4
comfort 3.97


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def new
+      @post = Post.find(params[:post_id])
+      @comment = Comment.new
+  end
+
+  def create
+      @post = Post.find(params[:post_id])
+      @comment = @post.comments.create(comment_params)
```
mattscilipoti, line 15: Recommend using bang method (`@post.comments.create!...`) or the conditional (`if @post.comments.create...`)

mattscilipoti:
:+1:

PR #100, opened 2015-11-11
Dt solution

dttruong:
comfort = 3
completeness = 3

I was having trouble pulling up my rails server after messing around with it.


nolds9:
:+1:  Great job on the first part! Missing the second part of the scribble hw, please submit that ASAP

PR #102, opened 2015-11-12
Kelly.log

kaustly:
First part of Scribble


app/views/articles/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1>Listing articles</h1>
+<%= link_to 'New article', new_article_path %>
+<table>
```
adambray, line 3: It's not a good idea to use tables, except for tabular data. Here I'd suggest using divs or article or section tags.

app/views/articles/_form.html.erb
```diff
@@ -0,0 +1,31 @@
...
+        <%= pluralize(@article.errors.count, "error") %> prohibited
+        this article from being saved:
+      </h2>
+      <ul>
+        <% @article.errors.full_messages.each do |msg| %>
+          <li><%= msg %></li>
+        <% end %>
+      </ul>
+    </div>
+  <% end %>
```
adambray, line 15: This format, and the use of tables in the index, make me think this was generated with `rails generate scaffold`, which isn't wrong per se, but I'd encourage you (if you did) to try and avoid using that generator as much as possible, since it's really important to understand how to build this code from scratch

PR #103, opened 2015-11-12
first part

matiamin:
comfort 3

completeness 3

Adam, I'm still working on it.


scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+  def index
+    if current_user
+      @posts = current_user.posts
+    else
+      @posts = Post.all
+    end
+  end
+
+  def create
+    @post = current_user.posts.create(post_params)
```
adambray, line 17: :+1:

scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  def destroy
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.find(params[:id])
+    @comment.destroy
+    redirect_to post_path(@post)
+  end
+
+  private
+    def comment_params
+      params.require(:comment).permit(:commenter, :body)
```
adambray, line 18: Main thing I'd suggest if you were to keep working on this is to set up the relationship between comments and users, just like you did with posts.

scribble/app/views/comments/_comment.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= comment.commenter %>
+</p>
+
+<p>
+  <strong>Comment:</strong>
+  <%= comment.body %>
+</p>
+
+<p>
+  <%= link_to "Delete Comment", [comment.post, comment],
```
adambray, line 12: Nice! I didn't know you could pass in an array of objects here to generate the nested 'show' path.

scribble/app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_for([@post, @post.comments.build]) do |f| %>
```
adambray, line 1: Nice!

scribble/app/views/posts/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<h1>Scribble Posts</h1>
+<%= link_to "New post", new_post_path%>
+
+<table>
```
adambray, line 4: It's best never to use a table for layout. These should be semantic tags like `<article>` or `<section>` or at least `<div>`s.

Tables should only be used for tabular data.

adambray:
Good job!

PR #105, opened 2015-11-12
rebecca solution scribble

rebeccaestes:
completeness: 4
comfort: 4

I couldn't get a timestamp to work, and the comments/blogs reorder themselves whenever you edit one so that the most recently changed entry is at the bottom (I could probably fix the latter if the timestamps had added to the database). Otherwise I think everything works.

Also, I didn't end up using the comment/show or comment/index views for anything, although I haven't deleted them, or removed them from comments_controller ...


ebirving:
:tada: You've made so much progress this week, Rebecca! 

You definitely can delete the views and controller actions you don't end up needing/using. In fact, doing so will make your code all the most readable. (Speaking of readability, _please_ make sure you're cleaning up your indentation before you submit assignments!)

I'm seeing the timestamps in the database just fine. Maybe you just haven't migrated/deployed/pushed the latest changes? As far as the ordering issue, the `sort_by` enumerable may be able to help you out...

app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,12 @@
+<% @comment[:blog_id] = @blog %>
+<% @comment[:user_id] = @user %>
+
+
+<%= form_for [@blog, @comment] do |f| %>
+
+ <p> <%= f.label :content %><br />
```
ebirving, line 7: HTML5 gives us a nifty `<fieldset>` tag for use where you have `<p>` tags in forms -- nice specificity without having to mess with extra selectors.

readme.md
```diff
@@ -8,31 +8,31 @@ with the best content all around the world. It is designed to be built, refined,
...
 - Implement nested routes.
 
-## Sessions (due Friday)
+## Sessions (due Friday) DONE
 
 - Add User model with authentication.
 - Associate posts with users.
 
-## Deploy (due Friday)
+## Deploy (due Friday) Do Thurs?
```
ebirving, line 37: Nice workflow technique! :smile_cat:

PR #107, opened 2015-11-13
hard but fun assignment

MohamedGood:
3 out of 5


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  def update
+    @post = Post.find(params[:id])
+    @post.update(artist_params)
+    redirect_to post_path(@post)
+  end
+
+  def destroy
+      @post = Post.find(params[:id])
+      @post.destroy
+      redirect_to post_path
```
ebirving, line 33: The post doesn't exist any more (yay successful delete action!), so trying to redirect to the `post_path` throws an error. What if you redirected to `posts_path` (the posts index page) instead?

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      @post = Post.find(params[:id])
+    end
+
+  def edit
+        @post = Post.find(params[:id])
+      end
+
+  def update
+    @post = Post.find(params[:id])
+    @post.update(artist_params)
```
ebirving, line 26: Oops! I think you meant `post_params`.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,49 @@
+class CommentsController < ApplicationController
+
+  def index
+    @comments = Comment.all
+  end
+
+  # new
+  def new
+    @post = Post.find(params[:artist_id])
```
ebirving, line 9: `:artist_id`! Be careful about replacing alll appearances of a variable when you reuse code from other apps.

ebirving:
Well done! :+1: This is really impressive progress in such a short amount of time! 

There _are_ a handful of syntax errors, some of which I point out in my line comments, that impede the functionality of the app. They're easy fixes, but they cause big problems for users. Be sure to test all of the endpoints in your apps before you ship. For Scribble, this would mean trying to create, read, update, and destroy both posts and comments to make sure everything works. Even if the errors you find end up being more difficult to solve, making note of bugs in your submission notes (or readme in the case of a deployed project), shows ownership and understanding of your code.

PR #111, opened 2015-11-13
Chris_Kibble_Friday

ckib16:
View, nested routes and helpers done

To do-
Devise & Deploy to Heroku

confidence - 3/5
comments - I feel like I need to memorize the controller logic/code. Does this get memorized at some point thru repetition, or are you always looking up this code for each action?


RobertAKARobin:
It gets memorized at some point... and then you start using controllers for things other then CRUD, and everything you had memorized goes out the window. :)

Remember there isn't a "right" way to do controllers. There's a generally-recommended way. If you get stuck you can certainly look it up.

Gemfile
```diff
@@ -0,0 +1,47 @@
+source 'https://rubygems.org'
+
+gem 'rails-erd'
+gem 'ruby-graphviz'
```
RobertAKARobin, line 4: Cool! Never seen these gems before.

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,11 @@
...
+
+<ul>
+  <% @posts.each do |post| %>
+  <li>
+    <%= link_to post.title, post_path(post) %>
+  </li>
+  <% end %>
+</ul>
+
+<%= link_to "New Post", new_post_path, class: "btn btn-primary" %>
```
RobertAKARobin, line 11: Yay, Bootstrappiness!

db/migrate/20151111030421_create_comments.rb
```diff
@@ -0,0 +1,9 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.text :body
+      t.references :post, index: true, foreign_key: true
```
RobertAKARobin, line 5: When you use `.references`, you actually don't need to do `index` and `foreign_key`. These lines are equivalent:

``` rb
t.references :post
t.integer :post_id, index: true, foreign_key: true
```

db/seeds.rb
```diff
@@ -0,0 +1,52 @@
...
+
+Comment.create({
+  body: "Here's my comment 1",
+  post_id: 2
+  })
+
+Comment.create({
+  body: "Here's my comment 2",
+  post_id: 2
+  })
```
RobertAKARobin, line 52: The trouble with this is that it's assuming what each post's ID will be. If you run the seed file more than once, this will be inaccurate. I recommend saving each post to a variable and then doing `@post1.comments.create()`

RobertAKARobin:
Did you manage to get this deployed? Remember to deploy early and deploy often!

Gemfile
ckib16, line 4: Yeah - I love rails-erd. One rake command (`rake erd`) and boom - ERD diagram done as PDF in root folder.

db/migrate/20151111030421_create_comments.rb
ckib16, line 5: Cool, thanks. Migrations are and area I'm working on.

Found a cool GUI-based app the imports/exports Rails migrations, but probably overkill for what we are doing - https://www.malcolmhardie.com/sqleditor/

db/seeds.rb
ckib16, line 52: I think I understand - I'll ask for more clarification on the code this week.

ckib16:
No - I moved on due to time-boxing and being behind schedule, but I have no questions on Heroku deployment.

PR #115, opened 2015-11-13
Tsolow scribble

solowt:
completeness: 5/5
comfort: 4/5
https://tsolowscribbler.herokuapp.com/posts

Plan to do a little work on this tomorrow morning with validations/flash messages and perhaps improve the way I implemented tags.

Tags do function, and everything else seems to be working.  Note: the seeds.rb file depends on having a user in the DB already.  If there is a user in the DB, it will work.  If not, it won't.  I'm unsure how to get around this since I don't know of a way to create a new user without using the devise form on the actual page.

Also, the latest version isn't pushed to heroku as of me writing this.  Also plan to do that in the morning, but the heroku version does work.

edit- Added some validations to prevent blank posts and to check for 404s.  Also pushed latest version to heroku.


app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+  #creates a new tag, also creates a category here if it doesn;t exist
+  def create
+    #uses this boolean to check for a tag already existing.  I'm sure there's way better
+    #methods for this
+    bool = "true"
+    @post = Post.find(params[:post_id])
+    #if the category doesn't already exist...
+    if !Category.find_by(descrip: params[:category])
+      new_cat = Category.create!(descrip: params[:category]) #create the new category using params
```
nolds9, line 16: How could we use strong_params to sanitize user input?

app/controllers/tags_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+  def show
+  end
+  def new
+  end
+  #creates a new tag, also creates a category here if it doesn;t exist
+  def create
+    #uses this boolean to check for a tag already existing.  I'm sure there's way better
+    #methods for this
+    bool = "true"
```
nolds9, line 12: we could use `.find_or_create_by` to dry up this code: http://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by

config/routes.rb
```diff
@@ -0,0 +1,9 @@
+Rails.application.routes.draw do
+  devise_for :users
+  root :to => "posts#index"
+  resources :categories
+  resources :posts do
+    resources :tags
```
nolds9, line 6: Thinking we do not need for tags to be a nested resource of posts. Or feel that we can limit the routes that are exposed. Something like `resources :tags, only: [:create, :destroy]`

app/controllers/categories_controller.rb
```diff
@@ -0,0 +1,18 @@
+class CategoriesController < ApplicationController
+  def index
```
nolds9, line 2: Again, here I would limit which action's are exposed or only include the ones we care about

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.new(comments_params)
+    if @post.save
+      flash[:notice] = "Comment successfully created."
+      redirect_to @post
+    else
+      flash[:alert] = "Comment creation failed, likely missing a needed field."
+      render :new
+    end
```
nolds9, line 30: :+1:

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    @comment.destroy
+    redirect_to post_path(@post)
+  end
+  #links to the edit form
+  def edit
+    @post = Post.find(params[:post_id])
+    @comment = Comment.find(params[:id])
+  end
+  #updates comments here, must be the owner of a post in order to update comments attached to that post
+  def update
```
nolds9, line 45: How are we checking if the user who left the comment, or is the original owner of the post, can be the one who deletes/updates comments

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
+class CommentsController < ApplicationController
+  # lists all comments, separate from the post. this is fairly useless
```
nolds9, line 2: Do we need some kind of authorization for these controller actions? Depending on your app's logic, who should be able to CRUD comments?

nolds9:
:+1:  Excellent job on this assignment. Please review inline comments for feedback, and let me know if you have any questions.

PR #116, opened 2015-11-13
initial commit: most functionality done... 

ghost:
Need to make the comments work and then add the style. Gotta make it pretty and stuff.


app/assets/stylesheets/articles.scss
```diff
@@ -0,0 +1,23 @@
...
+
+h1 {
+
+
+}
+
+p {
+
+
+}
```
mattscilipoti, line 23: ???

mattscilipoti:
:+1:  More commits!  Please make a conscious decision about what you commit and when.  You tend to use rails generators.  It would be really helpful if you committed just after generation, so we could see what effort is yours.  Also, you've committed a pile of files from tmp/cache, this dir should be in your gitignore file.

PR #117, opened 2015-11-13
Joe gz scribble solution part2

joe-gz:
comfort: 3.5-4/5
completeness: 4.5

I think i hit all of the deliverables, and I feel like I'm really close to the bonus.  I can add categories to posts, and click the link on that category to show the given post. But, I'm struggling to link multiple posts with one category. 
As per usual, I'm also a little behind with styling, and I would have liked to add in some flash stuff. But, I wanted to get the functionality down first.

https://scribble-joegz.herokuapp.com/


app/assets/stylesheets/application.css
```diff
@@ -0,0 +1,69 @@
...
+* You're free to add application-wide styles to this file and they'll appear at the bottom of the
+* compiled file so the styles you add here take precedence over styles defined in any styles
+* defined in the other CSS/SCSS files in this directory. It is generally better to create a new
+* file per style scope.
+*
+*= require_tree .
+*= require_self
+*/
+
+body {
```
ebirving, line 17: :+1:

app/assets/stylesheets/application.css
```diff
@@ -0,0 +1,69 @@
...
+}
+
+.post_input {
+  height: 20%;
+}
+
+.post_title {
+  padding-top: 2%;
+  padding-bottom: .5%;
+  font-family: 'Josefin Slab', sans-serif;
```
ebirving, line 35: You could eliminate all this repetition of 'Josefin Slab' by assigning font-family to body...

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,59 @@
...
+    @posts = Post.all
+  end
+
+  def index
+    if current_user
+    @posts = current_user.posts
+  else
+    @posts = Post.all
+  end
+  end
```
ebirving, line 15: I see what you're accomplishing with `allpost` and `index`, and your instinct to deviate from RESTful actions is dead-on, but the execution could be a bit more semantic. To me, `allpost` seems to serve the function we expect from a tradition `index`. What if you named `allpost` back to `index`, and gave your non-RESTful name to the action with the conditional? Something like `user_posts`? You could also look into filters if you want to get fancy...

ebirving:
:tada: Great work, Joe! The infrastructure for adding multiple posts to one category seems to be there. It sounds like you're looking for something like a dropdown of existing categories on the `form_for [@post,@category]`? 

A couple of overall things -- be careful about your indenting. I'm noticing it mainly in the controllers. And speaking on controllers, at this point you should feel free just delete any actions you're not using and don't anticipate needing. You're at the point now where you can start customizing the full RESTful template to fit the needs of your app.

I'm excited for your Project 2!

PR #118, opened 2015-11-13
Alexis_Scribble_Part2

ajohnson052:
Completeness: 4
Comfort: 3

I think this meets all of the requirements, but there is a lot more that I would like to do on it. The next step would be to ensure that a user can only edit their own posts. Is there some way to use before_action ... only: :edit for this or is it more appropriate to do some kind of conditional error handling within the edit/update method?


ajohnson052:
https://alexis-scribble.herokuapp.com/users/sign_in

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @comment = Comment.find(params[:id])
+  end
+
+  def edit
+    @comment = Comment.find(params[:id])
+  end
+
+  def update
+    @comment = Comment.find(params[:id])
+    @comment.update(comment_params)
```
jsm13, line 28: Could put a conditional here that current user is the comment's author else have a flash message only authors can edit comments

app/views/posts/show.html.erb
```diff
@@ -0,0 +1,19 @@
+<h2><%= @post.title %></h2>
+<h3><%= @post.user.email %></h3>
+<h3><%= @post.created_at.strftime("%m/%d/%y")%></h3>
+<div>
+  <p class="text"><%= @post.text %></p>
+  <%= link_to "Edit Post", edit_post_path(@post) %>
```
jsm13, line 6: Can have a conditional here that only shows the edit link if the current user is the author of the post

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    @post = Post.find(params[:id])
+  end
+
+  def edit
+    @post = Post.find(params[:id])
+  end
+
+  def update
+    @post = Post.find(params[:id])
+    @post.update(post_params)
```
jsm13, line 26: And a conditional here that makes sure the current user is the author before it makes an update

jsm13:
:+1: Looks great. To only let authors edit you can use conditionals with devises' current_user method. You just have to make sure you cover the possible ways to get to something. For example, you only want show edit links to authors, but also want to redirect users who aren't the author trying to navigate directly to the edit page's url. You'd also want to check before the update that the current_user (method provided by devise) is the author so that somebody wouldn't be able to make a PATCH request directly to the url using another tool.

PR #121, opened 2015-11-13
Matthewmurgia solution

murgia:
comfort:3
complete:4


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,53 @@
+class CommentsController < ApplicationController
+  # index
+  def index
+    @comment = Comment.all
+  end
```
andrewsunglaekim, line 5: do we need this action? At any point will you want to see all of the comments in your entire application? I think we'll only ever want to see comments associated with a post

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,42 @@
+class PostsController < ApplicationController
+  def index
```
andrewsunglaekim, line 2: interesting you want authentication restrictions on just the index route. Almost seems like CUD functionality should be protected and allow others to view whatever they want. But it totally depends on the domain you want for your scribble application.

andrewsunglaekim:
:+1:  Good job on the assignment matt.

PR #122, opened 2015-11-13
Andres solution 2

codedre:
completeness: 3
comfort: 3

Tried to add to much that didnt work, failed to commit often, could not successfully deploy


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def edit
+    @post = Post.find(params[:post_id])
+    @comment = Comment.find(params[:id])
+  end
+
+  def update
+    @post = Post.find(params[:post_id])
+    @comment = Comment.find(params[:id])
```
mattscilipoti, line 31: Recommend `@post.comments.find...`  This limits it to a comment for this post (even if someone hijacks the request.

mattscilipoti:
:+1:
I see 2 common issues.  
- The .gitignore file is missing. I assume you had to move the app up one dir.  Without this file, git will attempt to manage files it should not (log/_, tmp/_).
- You used `git add .`, immediately followed by `git commit -m...`, which caused you to commit a pile of temporary files, which are updated all the time.  I recommend a `git status` between those commends.  Or use `git commit`, without the `-m`, and write your comment in your editor.
  After add `.gitignore`, you can stop git from tracking these via `git rm`.

PR #124, opened 2015-11-13
Beauchamp Scribble Solution

beckybeauchamp1:
comfort_level: 3
completeness: 4

heroku: https://beauchamp-scribble.herokuapp.com/

I still could not fix the strong params for devise and allow name to be added as a parameter. Although, I read through the documentation for devise, I would really like to learn more information on user authentication and how I better manipulate what the devise gem


scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+  def new
+    @comment = Comment.new
+    @post = Post.find(params[:post_id])
+  end
+
+  def create
+    @post = Post.find(params[:post_id])
+    @post.comments.create!(comment_params.merge(user: current_user))
+    redirect_to post_path(@post, @user), notice:
+    "#{current_user.email} has left a new comment!"
```
adambray, line 14: I like how you broke this onto a separate line to keep the line length below 80, but I'd break it a bit differently according to common ruby style:

``` ruby
redirect_to post_path(@post, @user), 
  notice: "#{current_user.email} has left a new comment!"
```

notice the break after a comma, and the indentation to indicate the lines are really 1

scribble/app/controllers/home_controller.rb
```diff
@@ -0,0 +1,5 @@
+class HomeController < ApplicationController
+  def index
+    @recent_posts = Post.all.order(:created_at).limit(3)
```
adambray, line 3: :+1:

scribble/app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= form_for [@post, @comment], :html =>{ :class =>'form' } do |f| %>
```
adambray, line 1: Nice use of a partial!

scribble/app/views/users/confirmations/new.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>Resend confirmation instructions</h2>
```
adambray, line 1: I suspect the reason that you're having trouble with over-riding defaults here is that you need to have these views in `app/views/devise` instead of `app/views/users`, but I'm happy to pair with you on this at some point if you want.

adambray:
Fantastic job!  I really don't have a lot to add, there's so much here that's great!

PR #125, opened 2015-11-13
kaustly solution

kaustly:
comfort:3
completeness: 2


config/routes.rb
```diff
@@ -0,0 +1,63 @@
+Rails.application.routes.draw do
+  root 'welcome#index'
+  resources :articles do
+  resources :comments
+end
```
adambray, line 5: Make sure your indentation is correct, this should be indented one!

scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.create!(comment_params)
+    redirect_to @post
+  end
+
+  def destroy
+    @post = Post.find(params[:post_id])
+    @post = @post.comments.find(params[:id])
+    @comment.destroy
+    redirect_to (@post)
```
adambray, line 15: it's best not to put a space between a method and the parens with it's arguments. can lead to errors / bugs in some situations

scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+  # GET /posts/1/edit
+  def edit
+  end
+
+  # POST /posts
+  # POST /posts.json
+  def create
+    @post = Post.new(post_params)
+
+    respond_to do |format|
```
adambray, line 39: what is this doing?

PR #127, opened 2015-11-13
Jmoon solution2

mooniker:
Completeness: 3/5
Didn't get authentication working.

Comfort 3/5

The version working before I attempted authentication is at:
https://scribbble.herokuapp.com/


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,23 @@
+class CommentsController < ApplicationController
+
+  http_basic_authenticate_with name: "dhh", password: "secret", only: :destroy
```
andrewsunglaekim, line 3: waht does this do?

andrewsunglaekim, line 3: so this is kind of a quick way to do not so real authentication, with the devise gem we can streamline this functionality into a model.

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,41 @@
+class UsersController < ApplicationController
+
+  def sign_up
+  end
+
+  def sign_up!
```
andrewsunglaekim, line 6: i like that you're trying to hand roll a user auth solution. We need a way to not store any password literals into our database. We want to process passwords into some kind of hashing algorithm. I recommend the `BCrypt` Gem

app/models/post.rb
```diff
@@ -0,0 +1,5 @@
+class Post < ActiveRecord::Base
```
andrewsunglaekim, line 1: :+1:

app/controllers/users_controller.rb
andrewsunglaekim, line 6: one thing to note, is this code will severely conflict with the devise installation.

andrewsunglaekim:
I think alot of your issues with the auth portion stem from having your on handrolled auth solution as well as a devise solution. :+1:  great work on this assignment though

PR #129, opened 2015-11-13
Pengelly solution

jenpen:
comfort: 4
completeness: 2.5

still need users and to deploy. ran into issue with having created user table before devise gem. subsequently did some wrong things to resolve.


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def index
+    @posts = Post.all
+  end
+
+  def new
+    @post = Post.new
+  end
+
+  # Need to replace with new/save if statement
+  def create
```
andrewsunglaekim, line 12: put those validations to use! I'm glad you commented on it =D

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,42 @@
+class PostsController < ApplicationController
+  before_action :set_post, only: [:show, :edit, :update, :destroy]
```
andrewsunglaekim, line 2: cool, i like this `set_post` method

andrewsunglaekim:
Looks great, although the git diff had 531 changed files.. So i'm not entirely sure if I saw everything. Think you had more then a 2.5 in completeness for sure though. Devise can be tricky, especially if you started a handrolled solution. If I know i'm going to use devise in an application, it's generally one of the first things I mount. Great job though jen! :+1:

PR #133, opened 2015-11-13
Negar's Solution

neka-io:
Comfort Level: 3

Completion: 4
https://scribble-nk.herokuapp.com/

This week's been tough in terms of dedicating time to homework so this isn't as complete as I'd wanted it to be.


app/assets/stylesheets/application.css
```diff
@@ -0,0 +1,63 @@
...
+  text-decoration: none;
+}
+.clearfix:after {
+     visibility: hidden;
+     display: block;
+     font-size: 0;
+     content: " ";
+     clear: both;
+     height: 0;
+     }
```
jsm13, line 63: I've really liked using flexbox but this totally works. The app looks great

app/views/users/registrations/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <%= devise_error_messages! %>
+
+  <div class="field">
+    <%= f.label :username %><br />
+    <%= f.text_field :username, autofocus: true %>
+  </div>
+
+  <div class="field">
+    <%= f.label :email %><br />
+    <%= f.email_field :email, autofocus: true %>
```
jsm13, line 13: One is meant to style fields with css rather than break tags but this is exactly what I do. I'm trying to do better though

jsm13:
:+1: Looks good! Anything you've got questions on?

PR #134, opened 2015-11-13
Jordan brauner friday solution

jordanbrauner:
Comfort: 3
Completeness: 4.5

Comments: All features are in and working except for bonus and Heroku deployment. I was getting a failed push when I typed git push jordan-brauner-solution:master. Will look into asap so I can get it up before you look at it.


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def new
+    @post = Post.find(params[:post_id])
+    @comment = Comment.new
+  end
+
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = Comment.create!(comment_params.merge(post: @post))
```
amaseda, line 14: Have never tried doing it this way. I'm partial to the shorter `@post.comments.create( comment_params )` (especially because it's very readable).

app/views/layouts/application.html.erb
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>SUPER ROBOT</title>
+  <%= stylesheet_link_tag    'application', media: 'all', 'data-turbolinks-track' => true %>
+  <%= javascript_include_tag 'application', 'data-turbolinks-track' => true %>
+  <%= csrf_meta_tags %>
+  <link href='https://fonts.googleapis.com/css?family=Anton|Open+Sans:400,300,700,300italic,400italic,700italic' rel='stylesheet' type='text/css'>
```
amaseda, line 8: Rails might take care of this automatically, but in general I would place this link to a Google font before the stylesheet that uses it.

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<div id="posts">
+<% @posts.each do |post| %>
+  <h2 id="post-title"><%= link_to post.title, post %></h2>
+  <p id="post-body"><%= post.body %> </p>
```
amaseda, line 4: We only use `id`'s on a single element. While an `id` is technically reusable, that's what we use classes for.

amaseda:
Great job with this assignment! Let me know if you have any additional questions, but sounds like you've got a good grasp on Rails.

For future reference, I would include things like the `tmp` and `log` directories in your `.gitignore` file. Otherwise, your repo is going to have 1,000+ files in it. You can see how crazy that looks when you click `Files changed` on a pull request!

:+1:

PR #135, opened 2015-11-13
Scribble Homwork

qadriyyah119:
comfort level: 3.5
completeness: 3.5

I haven't deployed yet. I will do that before I leave GA. I was having issues with one of the functions. I also needed to format the size of the image on the header.


scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def show
+    @post = Post.find(params[:id])
+    render :show
+  end
+
+  def new
+    @post = Post.new
+    render :new
```
nolds9, line 13: We do not need these explicit renders, because Rails uses implicit rendering as long as we follow RESTful conventions!

scribble/app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,13 @@
+<%= form_for([@post, @post.comments.build]) do |f| %>
```
nolds9, line 1: Where is @post defined?
What about defining an instance variable for the new comment to be created?

scribble/app/views/posts/edit.html.erb
```diff
@@ -0,0 +1,5 @@
+<h1>Editing Post</h1>
+
+<%= render 'form' %>
+
+<%= link_to 'Back', posts_path(post) %>
```
nolds9, line 5: `<%= link_to "Back", post_path(@post) %>`
This would send us back to the show view for the post we are editing as long as `@post` is defined.

Otherwise you could use `posts_path` to go back to the index view, note `posts_path` does not need an argument.

scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,58 @@
+class PostsController < ApplicationController
+  def index
+    @post = Post.all
```
nolds9, line 3: Convention here is to use `@posts` because we are referring to a collection of `posts`

scribble/app/controllers/users_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    flash[:notice] = message
+    redirect_to action: :log_in
+  end
+
+  def sign_out
+    flash[:notice] = "Thank you for visiting!"
+    redirect_to "/log_in"
+  end
+
+end
```
nolds9, line 43: This all looks good, but a little confused why you did not choose to use the method which we learned in class (by using the gem Devise)?

scribble/app/models/user.rb
```diff
@@ -0,0 +1,5 @@
+class User < ActiveRecord::Base
+  validates :username, presence: true, uniqueness: true
+  # validates :password_digest, presence: true
+  has_secure_password
```
nolds9, line 4: Can we think of any associations our user model can be utilized for in our application?

nolds9:
:+1:  Good job on this. Please review my inline comments, and let me know if you have any questions. I will look to your submitted issue for deployed url, and final repo.  Main question is why you did not use the gem `devise` for your app's user authorization, and choose to hand-roll it?

qadriyyah119:
I used the handroll method by mistake. I went to the curriculum as guidance and didn't realize I was using the wrong one. Adam pointed that out on Friday during the lab. 

Sent from Outlook

On Sun, Nov 15, 2015 at 3:16 PM -0800, "Nicholas Olds" notifications@github.com wrote:

  Good job on this. Please review my inline comments, and let me know if you have any questions. I will look to your submitted issue for deployed url, and final repo.  Main question is why you did not use the gem devise for your app's user authorization, and choose to hand-roll it?

PR #145, opened 2016-03-16
my first commit. index, create, edit, and show pages

kairajohnson:
still working on it....

comfort at first.... 2/3
comfort now.... 3.5/4

question: what is the difference between having : before or after? Robin explained that : after requires => after but I have seen it without that. are they interchangeable or is there a way to distinguish when/how you should use each method of : before and after?


RobertAKARobin:
They're interchangeable! These two lines are exactly the same:

```
@artist.update(name: "Juan")
@artist.update(:name => "Juan")
```

```
{age: 42}
{:age => 42}
```

kairajohnson:
ok. not in anything i've seen in class, but on other resources I've seen for example length: 17 without =>. I'm assuming that was ok because it was an option within an action/method?

RobertAKARobin:
It doesn't have anything to do with what its function is -- it's just a personal preference, like using semicolons. The rules are:
- If you put the colon before the word, you need a `=>`
- If you put the colon after the word, you don't

kairajohnson:
ok. thanks

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def new
+    @post = Post.new
+  end
+
+
+  def create
+    @post = Post.create!(post_params)
+
+    @post.save
```
joe-gz, line 15: not necessary with .create!()

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+
+  def edit
+    @post = Post.find(params[:id])
+  end
+
+  def update
+    @post = Post.find(params[:id])
+    @postt.update(artist_params)
```
joe-gz, line 30: make sure you're not copying straight from Tunr.. this line has a spelling error, and is passing in artist_params

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,46 @@
+class PostsController < ApplicationController
+
+  def index
+    @post = Post.all
```
joe-gz, line 4: shouldn't this be plural? `@posts`

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def edit
+    @post = Post.find(params[:id])
+  end
+
+  def update
+    @post = Post.find(params[:id])
+    @postt.update(artist_params)
+
+    redirect_to posts_path(@post)
```
joe-gz, line 32: will this work? you're redirecting to `posts_path`, which assumes multiple posts (Id guess the index). but you're passing in a single post

joe-gz:
this is a good start Kaira! there were a few spots in your post controller that I left comments on.. I'd definitely try to tackle the comments at some point as well!

PR #153, opened 2016-03-17
Keith Prifte scribble hw

khenry10:
Comfort: 4 (eventually got the swing and felt pretty good.  stuck at comments form)
Completeness: 3.5


app/views/comments/new.html.erb
```diff
@@ -0,0 +1,12 @@
+<h2> New Comment </h2>
+
+<%= form_for @paths.comment do |f| %>
```
joe-gz, line 3: was this meant o be `@posts`?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,27 @@
+class CommentsController < ApplicationController
```
joe-gz, line 1: I think you'll probably need to define a post somewhere here to actually connect comments and posts?

joe-gz:
Nice job Keith! looks like you're missing an option to edit posts or comments. Let me know if you have any questions about that!

PR #155, opened 2016-03-18
Maureen Scribble

maureenv:
comfort level: 3
completeness:3 

I'm really happy that I'm able to create a new post on my site. I didn't get as far as to create comments but if I had more time I probably could have been able to. 

Also, here is the ERD: 

POSTS
- title
  -author
  -author_email
  -body
  -image_url

COMMENTS (belong to Posts)
-name
-email
-comment


mattscilipoti:
Your comments sound great.  That's what I'm hoping for.

If you look at your changed files, you only see "Subproject commit..."  This means you ran `git init` on the dir above your project, instead of in it.  Please ask an instructor to help you correct this.

PR #156, opened 2016-03-18
Anissa solution

iamanissa:
Comfort score: 3 out of 5
Completeness score: 3 out of 5
Any comments: I feel like I'm getting some understanding, but I was having issues when trying to show my comments in the post show view and understanding variables in that context.


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
+class CommentsController < ApplicationController
+    def index
+        @post = Post.find(params[:post_id])
+        @comments = @post.comments.all
+        redirect_to post_path(@post)
```
amaseda: So when somebody wants to view all the posts, you want to direct them to a specific post page (i.e., `show` instead of `index`)

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+        @post = Post.find(params[:post_id])
+        @comment = Comment.find(params[:id])
+        @comment.destroy
+
+        redirect_to post_comments_path( @post )
+    end
+
+    private
+    def comment_params
+        params.require(:comment).permit(:response, :created_at, :updated_at, :post_id)
```
amaseda, line 54: `created_at` and `updated_at` are timestamps, so they'll be generating automatically absent of user input.

Do you need to include `post_id` in strong params? Isn't that something you can assign regardless of user input?

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        @post = Post.find(params[:id])
+        @post.destroy
+        redirect_to posts_path
+    end
+
+
+    private
+    def post_params
+        params.require(:post).permit(:title, :content, :created_at, :updated_at)
+    end
```
amaseda, line 39: Similar to my note in the comments controller: do you need `created_at` and `updated_at` in here?

app/models/post.rb
```diff
@@ -0,0 +1,6 @@
+class Post < ActiveRecord::Base
+    has_many :comments, dependent: :destroy
+    validates_presence_of :title
+    validates_presence_of :content
```
amaseda, line 4: :+1: validations

app/views/comments/edit.html.erb
```diff
@@ -0,0 +1,13 @@
+<h2>Edit Your Comment</h2>
+
+<%= form_for [@post, @comment] do |f|%>
+
+    <%=  f.label :response%>
+    <%=  f.text_area :response, :cols => "150", :rows => "10 "%>
+</br>
```
amaseda, line 7: `<br>`'s are no-no's! Try using CSS to take care of this behavior.

app/views/comments/new.html.erb
```diff
@@ -0,0 +1,8 @@
+<h2>Create New Comment</h2>
+
+<%= form_for [@post, @comment] do |f|%>
+    <%=  f.label :response%>
+    <%=  f.text_area :response, :cols => "150", :rows => "10 "%>
+</br>
+    <%= f.submit "Submit"%>
+<% end %>
```
amaseda, line 8: Your comment `new` and `edit` forms look identical. Could you use a partial to refactor this so that you only need to use one .erb file?

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,14 @@
...
+    <% @posts.each do |post|%>
+        <h2>
+            <%= link_to post.title, post_path(post) %>
+        </h2>
+        <p>Created At: <%= post.created_at%></p>
+        <p><%= post.content%></p>
+    <% end %>
+    <% if @posts.size == 0%>
+        <em>No Posts Found. Try Adding a Post.</em>
+    <% end %>
```
amaseda, line 13: Doesn't make much of a difference in this case, but do you want the `each` statement to be run if `@posts.size == 0` is true? How could you use a conditional to implement this?

amaseda:
:+1: Good job with this. Let me know if you have any additional questions/comments.

PR #159, opened 2016-03-18
Lyndsey's Solution

LNewsome:
Not comfortable.
Not finished.


app/controllers/comments_controllers.rb
```diff
@@ -14,13 +14,13 @@ def show
   end
 
   def update
-    @comment =
-    @comment =
+    @comment = Comment.find(params[:id])
+    @comment.update(comment_params)
```
mattscilipoti, line 18: Looking good.  Go find the requested comment.  Then update that comment.  "comment_params" should be a method that provides the "strong params" for the comment - it manages/identifies what params that can be passed to the `@comment.update` method.  You need to create the "comment_params" methods for this controller.

db/migrate/20160316134241_change_column_in_comment.rb
```diff
@@ -0,0 +1,4 @@
+class ChangeColumnInComment < ActiveRecord::Migration
```
mattscilipoti, line 1: These migrations do not have any changes in them.  When run, they won't change the database schema.

mattscilipoti:
Many things are close, but not quite.

PR #160, opened 2016-03-18
Jeffrey Eaton solution

jeffreyeaton76:
Completeness: 4.8
Comfort level: 5
Has full CRUD for posts but not for comments. I also still have anchor tags on the posts#index page but I have link_tos everywhere else (just ran out of time). 

ERD below is for some magical version that I thought would also have users and tags. 

img_1985


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,34 @@
+class CommentsController < ApplicationController
+
+  def index
+  end
+
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.create!(comment_params)
+    redirect_to post_path(@post)
```
joe-gz, line 9: can be shorted to simply `redirect_to @post`

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>My Blue Blog</h1>
+<p><a href="/posts/new">Create a Post!</a></p>
+  <div class="big-box">
+  <% @posts.reverse_each do |p| %>
+  <div class="post_container">
+    <p class="title"> <a href="/posts/<%= p.id %>"> <%= p.title %> </a></p>
```
joe-gz, line 6: Saw you noted it, but a link_to is preferred here!

joe-gz:
Nice job Jeff! let me know if you have any specific questions. Looks like you were close with CRUD for comments, is there anything specific you felt was missing?

PR #162, opened 2016-03-18
Andrew sm solution

arsm800:
comfort: 3.5
completeness: 3.5

I still have yet to set everything up for my "create" and "new" and "update" methods and html.erbs.


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+    @post = Post.find(params[:id])
+  end
+
+  def new
+    @post = Post.new
+  end
+
+  def edit
+    @post = Post.find(params[:id])
+  end
```
amaseda, line 18: You're going to need `create` and `update` controller actions if `new` and `edit` are going to serve a purpose, respectively. Guessing you're implementing this in your project but let me know if you have any questions about this.

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,5 @@
+
+    <header>Posts!</header>
+    <% @posts.each do |post|%>
+      <p><%= post.post_title %></p>
```
amaseda, line 4: It'd be helpful for the user if these post titles were links. Could use a link helper here.

config/routes.rb
```diff
@@ -0,0 +1,64 @@
+Rails.application.routes.draw do
+
+root to: 'posts#index'
+
+resources :posts
+resources :comments
```
amaseda, line 6: Based on how your routes and controller are currently set up, how would you go about accessing a comment(s) that belongs to a particular post?

db/migrate/20160315014311_create_posts.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePosts < ActiveRecord::Migration
+  def change
+    create_table :posts do |t|
+      t.string :post_title
+      t.string :post_datestamp
+      t.string :post_text
```
amaseda, line 6: Since this is a `posts` table, do you need to indicate `post_` in attribute names?

db/migrate/20160315014311_create_posts.rb
```diff
@@ -0,0 +1,12 @@
...
+      t.string :tag_title
+      t.string :tag_text
```
amaseda, line 8: This is an advanced topic, but `tags` are a classic example of a many-to-many relationship (and thus are usually not indicated as attributes in this way). Let me know if you'd like to talk about this.

db/migrate/20160315014321_create_comments.rb
```diff
@@ -0,0 +1,9 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :timestamp
+      t.string :comment_text
```
amaseda, line 5: Same note as your `create_posts` migration: do you need `comment_` in the attribute name?

db/migrate/20160315014321_create_comments.rb
```diff
@@ -0,0 +1,9 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :timestamp
```
amaseda, line 4: `timestamp` should be plural.

amaseda:
:+1: Good start with this. I imagine you're doing this with your project, but encourage you to implement full CRUD functionality here. Let me know if you have any questions/comments about that (or anything else).

PR #164, opened 2016-03-18
Alexa Klein scribble solution

aklein916:
completeness-- 3.5 
comfort level-- 4

I ended up dropping the database tonight and starting over, I think I originally made the table more complicated than necessary. I need more practice with the helpers and adding multiple users, which I look forward to spending time working on during the project next week.


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,21 @@
+class PostsController < ApplicationController
+  def index
+    @posts = Post.all
+  end
+
+  def new
+
+  end
```
amaseda, line 8: I'm guessing you're implementing this in your project, but encourage you to implement new/create functionality here if you need more practice with it.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,16 @@
+class CommentsController < ApplicationController
+  def index
+    @comments = Comment.all
+  end
+
+  def new
+    @comment = Comment.new 
+  end
```
amaseda, line 8: `new` can't do anything if it doesn't have a corresponding `create` controller action!

app/views/comments/new.html.erb
```diff
@@ -0,0 +1,12 @@
+
+<h3>Add a Comment!</h3>
+<%= form_for @comment do |f| %>
+  <%= f.label :name %>
+  <%= f.text_field :commentor %>
```
amaseda, line 5: Matching labels and text fields (or other types of inputs) should have matching names (e.g., both should be called `:name` or `:commentor`).

app/views/comments/new.html.erb
```diff
@@ -0,0 +1,12 @@
...
+
+  <%= f.label :comment %>
+  <%= f.text_area :body %>
```
amaseda, line 8: Same as above. Both should be either `:comment` or `:body`.

config/routes.rb
```diff
@@ -0,0 +1,19 @@
+Scribble::Application.routes.draw do
+  root "posts#index"
+  resources :posts, only: [:index, :show, :new]
+  resources :comments, only: [:index, :show, :new]
```
amaseda, line 4: So how do I go about accessing a comment(s) that belongs to a particular post? Can we accomplish that with the way your routes and controllers are currently set up?

amaseda:
:+1: Good start with this. Again, you're probably implementing this in your project, but encourage you to build out full CRUD functionality here.

Let me know if you have additional questions/comments.

PR #165, opened 2016-03-18
Homework Submission

Ceisen9:
Had a lot of trouble integrating it with actuals users. Note this takes place after the second commit.


jshawl:
:+1: wow awesome job getting started with integrating users! There will be a mini-lesson on Wednesday about a gem called devise which makes this simpler - come by!

PR #166, opened 2016-03-18
Adrian wyatt scribble solution

Wyatt23:
It's a lot to grasp but I'm heading there! I love this Rails thing - it's hard but great! 

so far I have a very pretty postable blog but still need to work on the edit functionality, and several others. The CSS is an ugly beast codewise but I'll refactor it later. I had fun with this! 

comfort level: 4 
completeness: 4 - timeboxed out


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def new
+    @comment = Comment.new
+  end
+
+private
+  def comment_params
+  params.require(:comment).permit(:content)
+  end
```
amaseda, line 13: Indentation is helpful esp. when dealing with private methods.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    @post = Post.create(post_params)
+    redirect_to posts_path
+  end
+
+  def show
+    @post = Post.find(params[:id])
+  end
+
+  def edit
+  end
```
amaseda, line 21: **UPDATE:** Noticing your PR note now. Happy to comment on this if/when you work on it further!

Are you using this for anything?

app/models/post.rb
```diff
@@ -0,0 +1,5 @@
+class Post < ActiveRecord::Base
+  validates_presence_of :post_content
```
amaseda, line 2: :+1: validations

app/views/comments/index.html.erb
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
+<% @comments.each do |comment| %>
+<div class="comment">
```
amaseda, line 3: Indentation in .erb files is important too!

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,9 @@
+
+<% @posts.each do |post| %>
+<p class="ptitle"><%= post.title%></p>
+<p class="pauthor"><%= post.author%></p>
+<div class="go_postal">
+  <p class="pcontent"><%= post.post_content[0,700]%>...(click below for full post)</p>
+  <p><%= link_to"see more",post_path(post),:class => "a"%></p>
```
amaseda, line 7: I would avoid using class names like "a", which in your stylesheet could be confused for an anchor (`<a>`) tag.

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,9 @@
+
+<% @posts.each do |post| %>
+<p class="ptitle"><%= post.title%></p>
+<p class="pauthor"><%= post.author%></p>
+<div class="go_postal">
+  <p class="pcontent"><%= post.post_content[0,700]%>...(click below for full post)</p>
```
amaseda, line 6: Cool way to preview a post!

app/views/posts/new.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= form_for @post, html: {class:"new_post_form"} do |f| %>
+<div>
+  <%= f.label :title,:class => "label" %>
+  <%= f.text_field :title, :class => "title_box input" %>
```
amaseda, line 4: Is this `input` class something you apply to all `<input>` tags? If so, do you need the class? If not, perhaps there's a more semantic class name you could use to differentiate the class name from the tag name.

db/migrate/20160315035541_create_users.rb
```diff
@@ -0,0 +1,8 @@
+class CreateUsers < ActiveRecord::Migration
+  def change
+    create_table :users do |t|
+      t.text :content
+      t.timestamps null: false
```
amaseda, line 5: Because timestamps are automatically generated when creating/updating a table row, not sure if this needs to be `null: false`. Doesn't hurt though!

amaseda:
:+1: Good job with this! Let me know if you have any additional questions/comments.

PR #168, opened 2016-03-18
confused still

Crunchwrap78:
comfortness level 3

making migrations is simple
it's when it comes to the controllers and routes that's tripping me up. it's a bit hard without the scaffolding like in .NET, but I can see why rails is better.


jshawl:
:+1: Let me know if you have more questions about routing and controllers. found this to be helpful for me - http://guides.rubyonrails.org/routing.html

PR #169, opened 2016-03-18
Winnies solution

winishahines:
comfort: 4
completeness: 3

I made posts and comments but comments cannot be deleted. I also didn't get a change to setup user sign in though users can enter a name when commenting. The page is not very pretty since I didn't get to the styling part.

Scribble ERD listed here:
scribble erd


RobertAKARobin:
:white_check_mark: Signing in was a bonus anyway. This looks really good! Is there anything specific on which you'd like feedback?

PR #170, opened 2016-03-18
Farishtas solution

LaFrish:
Completeness 3
Comprehension: 4

The site was working when I had the post#index with just the link of the blog posts but not working when I was trying to get the post body to work. grrrr. But I think I got the rest of the things to down okish.


app/views/posts/index.html.erb
```diff
@@ -0,0 +1,10 @@
+<h2><%=link_to "Create New Post", new_post_path %></h2>
+
+  <article>
+    <%= @posts.each do |post| %>
+    <h2><%= link_to post.title, post %></h2>
+    <p>
+    <%= post_content, posts_path %>
```
jshawl, line 7: shouldn't this be `post.content`?

jshawl:
:+1: Let me know if you'd like to fix this issue together! otherwise, looks good!

PR #173, opened 2016-03-18
Matt's Solution

MattLeeStevens:
Completeness: 4
Comfort: 4

Starting to get the hang of the MVC framework stuff and how the different parts of the app work with each other.  I think I got all the CRUD stuff working for posts and comments.  The nested routes make things kind of tricky when it comes to knowing which path to use and which controller actions are being referenced for different views.  For a while I was using button_to when I should have been using link_to.  Apparently button_to can’t make get requests.  Building an app up up from scratch was good practice and I learned a lot working on this.


mattscilipoti:
:+1:  Glad it's coming together.

> button_to can’t make get requests.
> In general, the web expectation is that buttons are for changing data, links are for viewing.  You _can_ change the http method of a button to get, but you usually don't want to.

PR #174, opened 2016-03-18
charleshwang_solution

ghost:
comfort: 3/5

completeness: 3/5

This HW (along with the Hogwarts lab) helped with understanding routes and helpers but I definitely need more practice. I'm still having a little trouble with nested routes and which arguments to pass in when using certain helper methods.

ERD:

POSTS: *tite, *date, *content (has many comments)

COMMENTS: *comment (belongs to post)


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,14 @@
+class CommentsController < ApplicationController
+
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.create(comment_params)
```
beckybeauchamp1, line 5: Great job here with your Comments Controller! Nice job keeping it "skinny" and only defining methods(the create method) that you are using in your app

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+  end
+
+  def new
+    puts "This is a new post"
+    @post = Post.new
+  end
+
+  def create
+    puts params
+    puts "New Post Created"
```
beckybeauchamp1, line 18: Totally fine to do "puts" or use `pry` when your debugging, just remember to take it out after!

config/routes.rb
```diff
@@ -0,0 +1,27 @@
+Rails.application.routes.draw do
+  root to: 'posts#index'
+
+  resources :posts do
+    resources :comments
```
beckybeauchamp1, line 5: :+1:

app/views/posts/show.html.erb
```diff
@@ -0,0 +1,26 @@
...
+<h4><%= @post.date %></h4>
+<%= @post.content %>
+
+<h4><%= link_to "EDIT", edit_post_path(@post) %></h4>
+
+<h2>Comments</h2>
+<% @post.comments.each do |c| %>
+  <p>
+    Comment:
+    <%= c.comment %>
```
beckybeauchamp1, line 11: :+1:

db/migrate/20160314175436_create_comments.rb
```diff
@@ -0,0 +1,8 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :comment
+      t.belongs_to :post
```
beckybeauchamp1, line 5: I would change this column in your posts table from `:comment` to maybe `:content` or `:text` so its a little more semantic

beckybeauchamp1:
Hi Charles,

Good job on Scribble! Please let me know if you still have any further questions about nested routes, I'm happy to go over it with you!

PR #175, opened 2016-03-18
Jblakeman solution

jblakeman:
Comfort level: 3.5 and rising


app/views/posts/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>All posts</h1>
+<%= link_to "Create New Post", new_post_path %>
+<% @posts.each do |p| %>
+    <h2><%= link_to p.title, post_path(p) %></h2>
```
joe-gz, line 4: can be shorted to `link_to p.title, p`

joe-gz:
Nice start here John! Let me know if you have specific questions. I know last week was tough, so I'm happy to run through whatever you need

PR #183, opened 2016-04-19
zaneh_solution

zanehmw:
comfort: 4
complete: 5


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def destroy
+    # @post = Post.find(params[:post_id])
+    @comment = Comment.find(params[:id])
+    @comment.destroy
+    redirect_to @comment.post
+  end
+
+  private
+    def comment_params
```
adambray, line 35: This method isn't being used...

app/views/posts/show.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    <%= link_to '(Delete)', comment_path(cmt),
+                   method: :delete,
+                   data: { confirm: 'Are you sure?' } %>
+    <%= link_to '(Edit)', edit_comment_path(cmt) %>
+  </div>
+<% end %>
+</div>
+
+
+<%= form_for Comment.new do |f| %>
```
adambray, line 17: Why not use the form partial here?

adambray:
👍

PR #184, opened 2016-04-20
Scribble: Wes_solution 

wesareed:
comfort_level: 4
completeness: 4

would like to discuss adding validation for comments...


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,19 @@
+class CommentsController < ApplicationController
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.create(comment_params)
```
adambray, line 4: Your validations for comments are correct, but you'd need to do something here to handle a failed create, like:

```
@comment = @post.comments.build(comment_params) # build is like new, but for when you are doing it through an association
if @comment.save
  redirect_to @post
else
  render @post #render the post page which includes the comment form 
end
```

adambray:
👍

PR #192, opened 2016-04-20
James reichard solution

superbuggy:
Completeness: 3.5
comfort_level: 4


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def create
+    @comment = Comment.new(comment_params)
+    if @comment.save
+      redirect_to post_comment_path, notice: "Comment successfully created!"
+    else
+      render 'new'
+    end
+  end
```
andrewsunglaekim, line 18: this is fine, but normally we probably would build a comment off of the post.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    else
+      render 'edit'
+    end
+  end
+
+  def destroy
+    @post = Post.find(params[:id])
+    @post.destroy
+
+    redirect_to posts_path, notice: "Comment successfully destroyed!"
```
andrewsunglaekim, line 42: check flash for this action and the `save` action

app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,29 @@
...
+</p>
+
+<p>
+<%= f.label :commenter %>
+<%= f.text_field :commenter %>
+</p>
+
+<%= f.submit %>
+
+<% end %>
```
andrewsunglaekim, line 29: was wondering if this worked. It's a form for posts but looks like commenter fields.

andrewsunglaekim:
Good work on this assignment. 👍

PR #193, opened 2016-04-20
initial scribble commit

jball23:
complete: 4
comfort: 2

frustrating!


jball23:
I undid a lot of what I had done on my initial commit to try and make it look better and use associations, began refactoring it, couldn't get it back to where I needed it to be and I don't have the energy/time to get it back and I want to get into project mode.  Sorry for the incompleteness!!

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,23 @@
+class CommentsController < ApplicationController
+
+def new
+  @comment = Comment.new
+end
+
+def create
```
andrewsunglaekim, line 7: how will it get the post id?

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,51 @@
+class PostsController < ApplicationController
+
+  def index
+    @posts = Post.all
+    @comments = Comment.all
```
andrewsunglaekim, line 5: do you want all of the comments here? how will you know which ones are for which post? In the view when we iterate through posts we can do something with all it's comments:

``` ruby
@posts.each do |post|
   post.comments.each do |comment|
    # does somethings with each comment
  end
end
```

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  end
+
+  def show
+    @posts = Post.all
```
andrewsunglaekim, line 9: i dont think you need all of the posts for the show action.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def update
+    @post = Post.find(params[:id])
+    @post.update(post_params)
+    redirect_to @post
+  end
+
+  def create
+    @post = Post.create(post_params)
+    @comment = Comment.create(@comment)
```
andrewsunglaekim, line 32: as a matter of separation of concerns and single responsibility. The create action should only create 1 model, the model the controller is based around. In this case, the posts controller should only create posts.

app/views/comments/_form.html.erb
```diff
@@ -0,0 +1,12 @@
...
+<div class="comment_form">
+  <%= form_for [@post, @comment] do |f| %>
+    <p><%= f.label :author %></p>
+    <p><%= f.text_field :author %></p>
+    <p><%= f.label :body %></p>
+    <p><%= f.text_area :body %></p>
+    <p><%= f.hidden_field @post.id %></p>
+       <%= f.submit %>
+  <% end %>
+</div>
```
andrewsunglaekim, line 12: 👍

app/views/posts/_post.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="all_posts">
+  <% @posts.each do |post| %>
+  <p><%= link_to post.title, post_path(post) %></p>
+  <p>Written By: <%= post.author %></p>
+  <p><%= post.body %></p>
+  <% end %>
+</div>
```
andrewsunglaekim, line 7: 👍

andrewsunglaekim:
Your close in a lot of ways. I would take another look at this assignment when you get a chance. Additionally, I would run through the Ruby on Rails Guides tutorial to get that muscle memory down. I think i did it something like 6 times by the end of the course. It's very helpful with regard to seeing all the moving pieces fit together.

PR #195, opened 2016-04-21
completed restful routes for posts and comments

marygriffus:
Comfort: 4 - not totally happy with tags/categories
Completeness: 5


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,27 @@
+class CommentsController < ApplicationController
+  def create
+    @comment = Comment.create!(comment_params)
```
adambray, line 3: Suggest not using `!` methods in controllers, as they lead to a generic error page if the request fails. Instead, using an `if` statement (like we did in the validations lesson) with either a redirect + flash message or re-rendering the form w/ error messages leads to a nicer UX.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  def show
+    @post = Post.find(params[:id])
+    @comment = Comment.new
+    @comments = @post.comments
+  end
+
+  def destroy
+    @post = Post.find(params[:id])
+    @post.destroy
+    redirect_to posts_path()
```
adambray, line 34: Just a note: I put in parens sometimes to remind everyone that path helpers are really methods, and some people do use them all the time, but the most common convention in rails is to omit them if you're not passing anything in.

(Totally a matter of style though!)

app/views/comments/edit.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= form_for @comment do |c| %>
+  <%= c.label :title %>
+  <%= c.text_field :title %><br>
+  <%= c.label :author %>
+  <%= c.text_field :author %><br>
+  <%= c.label :body %>
+  <%= c.text_field :body %><br>
+  <%= c.hidden_field :post_id%>
```
adambray, line 8: 👍

app/views/comments/edit.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= form_for @comment do |c| %>
+  <%= c.label :title %>
+  <%= c.text_field :title %><br>
```
adambray, line 3: `<br>` tags are usually not the best way to apply style. I'd suggest wrapping each pair of label and input in it's own div.

db/migrate/20160418232042_create_comments.rb
```diff
@@ -0,0 +1,12 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.string :body
+      t.string :author
+      t.string :title
+      t.integer :post_id
```
adambray, line 7: This is totally fine, and the way we showed everyone... but there is a slightly better way:

```
t.references :post
```

That still creates an integer post_id column, but it also adds an index and a foreign key constraint, so you can't create a comment that doesn't have a post_id that actually corresponds to a post in the posts table.

adambray:
👍

PR #199, opened 2016-06-30
timeboxed

AT89:
edits not working properly, James couldn't help me with it either. Not sure whats wrong with it; I tried other methods such as using patch but it just broke the whole server until I couldnt figure out how to even revert it back. So i had to do a hard git reset.

frustrated and its 8pm on wednesday, I am ending it to save my sanity/getting burned out;spent around 6-8hours total on it.

--code breaks when i do the nesting so i left comments nesting out

comfort level 3/5
completeness 2/5

Hoping that I can learn more on Rails with the project..


scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,22 @@
+class CommentsController < ApplicationController
+
+  def show
+    @comment = Comment.find(params[:id])
+  end
+
+  def new
+    @comment = comment.new
```
nolds9, line 8: think you need to instance variables defined - missing one for the `Post` this new comment should be associated to

scribble/app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  end
+
+  def create
+    @post = Post.find(params[:post_id])
+    @comment = @post.comments.create(comment_params)
+    redirect_to @post
+  end
+
```
nolds9, line 16: edit / update for comments?

scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  def update
+    @post = Post.find(params[:id])
+    @post.update(post_params)
+      redirect_to @post
+  end
+
+  def create
+    @post = Post.create(post_params)
+
+    if @post.save
```
nolds9, line 28: this line is not being used because of the `.create` above I think.

scribble/app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def update
+    @post = Post.find(params[:id])
+    @post.update(post_params)
+      redirect_to @post
+  end
+
+  def create
+    @post = Post.create(post_params)
```
nolds9, line 26: probably want `.new` here since you are checking to see if it saves below

scribble/app/models/post.rb
```diff
@@ -0,0 +1,5 @@
+class Post < ActiveRecord::Base
+  has_many :comments
+  validates :topic, presence: true,
+                  length: { minimum: 5 }
```
nolds9, line 4: 👍

scribble/app/views/posts/edit.html.erb
```diff
@@ -0,0 +1,27 @@
+<h1> Edit post</h1>
+
+<%= form_for :post, url: posts_path do |post| %>
```
nolds9, line 3: Believe this is the wrong syntax for `form_for`... what about something like:

```
<%= form_for @post do | f | %>
   # all fields here
<% end %>
```

scribble/app/views/posts/show.html.erb
```diff
@@ -0,0 +1,40 @@
...
+  </p>
+
+  <p>
+    <strong>Comment:</strong>
+    <%= comment.text %>
+  </p>
+<% end %>
+
+<h2>Add a comment:</h2>
+<%= form_for([@post, @post.comments.build]) do |f| %>
```
nolds9, line 26: need to define `@post` in controller...

nolds9:
:+1: Good job on this, please review my comments and let me know if you have any questions. Would encourage you to come back to this and implement the rest of the outlined features.

PR #201, opened 2016-06-30
Gandy Young Scribble HW

gandyland:
Completeness: 2 or 3
Comfort: 3

While I felt kind of comfortable figuring out and writing some of the code I had trouble getting any kind of helpful feed back from the error codes to further my code. Overall I should have started this assignment earlier and probably would have had more luck.

Issues I had with this assignment:
I couldn’t figure out how to get the migrations to work.
Had a lot of trouble with the error codes and working them out. I think this was because I did some things out of order when trying to add code and because of my migration problem my database wasn’t responding correctly.


db/schema.rb
```diff
@@ -0,0 +1,19 @@
...
+# Note that this schema.rb definition is the authoritative source for your
+# database schema. If you need to create the application database on another
+# system, you should be using db:schema:load, not running all the migrations
+# from scratch. The latter is a flawed and unsustainable approach (the more migrations
+# you'll amass, the slower it'll run and the greater likelihood for issues).
+#
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 0) do
+
```
nolds9, line 15: Looks like there is no database structure defined...In order to set up a define a schema, you need to create and then run some migrations.Recommend reviewing http://guides.rubyonrails.org/active_record_migrations.html#migration-overview

app/models/post.rb
```diff
@@ -0,0 +1,4 @@
+class Post < ActiveRecord::Base
+  belongs_user :user
```
nolds9, line 2: `belongs_to :user`

app/models/user.rb
```diff
@@ -0,0 +1,6 @@
+class User < ActiveRecord::Base
+  devise :database_authenticatable, :registerable,
```
nolds9, line 2: looks like you are using Devise, we haven't talked about this yet, but can be a little difficult to configure

app/views/posts/index.html.erb
```diff
@@ -0,0 +1 @@
+
```
nolds9, line 1: How might you display all posts?

nolds9:
:+1: Thanks for submitting what you have. Definitely would recommend reviewing the role and syntax for migrations, as they are very important. I would encourage you to come back to this and implement the rest of the outlined features.

PR #205, opened 2016-06-30
Adina - Scribble

adina-pollack:
Level of comfort: 4.5
completeness: 5ish but I could probably shorten some of the controller methods


db/migrate/20160629205227_create_categories.rb
```diff
@@ -0,0 +1,8 @@
+class CreateCategories < ActiveRecord::Migration
+  def change
+    create_table :categories do |t|
+      t.string :name
+      t.integer :post_id
```
nolds9, line 5: don't think this foreign key is necessary

db/schema.rb
```diff
@@ -0,0 +1,44 @@
...
+# It's strongly recommended that you check this file into your version control system.
+
+ActiveRecord::Schema.define(version: 20160629205227) do
+
+  # These are extensions that must be enabled in order to support this database
+  enable_extension "plpgsql"
+
+  create_table "categories", force: :cascade do |t|
+    t.string  "name"
+    t.integer "post_id"
```
nolds9, line 21: Don't think this foreign key is necessary http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association

config/routes.rb
```diff
@@ -0,0 +1,61 @@
+Rails.application.routes.draw do
+  resources :posts do
+    resources :comments
+  end
+
```
nolds9, line 5: what might the routes look like if we wanted to implement tags/categories

nolds9:
:+1: Great job on this assignment. Would love to see you work out the rest of the functionality for `Categories` and `Tags` Let me know if you have any questions

PR #206, opened 2016-06-30
Scribble Tyson Green

Thaisuhn:
Comfort 3.48/5
Complete 3/5


app/models/post.rb
```diff
@@ -0,0 +1,2 @@
+class Post < ActiveRecord::Base
```
jshawl, line 1: `has_many :comments`

app/models/comment.rb
```diff
@@ -0,0 +1,2 @@
+class Comment < ActiveRecord::Base
```
jshawl, line 1: `belongs_to :post`

jshawl:
👍 Posts look great! Keep at it w/ the comments and let me know if you'd like to pair on this functionality!

PR #208, opened 2016-06-30
Dave C Scribble

dcordz:
Finished, I think, maybe? Started on the bonus but was too over it

Completeness: unknown, 4/5?
Comfort 2/5


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,48 @@
+class CommentsController < ApplicationController
+  def index
+    @comments = Comment.all(post: @post)
```
amaseda, line 3: You need to define `@post` in this controller action before you can reference it. I would do that on the prior line.

You would then be able to write out something like `@comments = @post.comments`

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    @comment.destroy
+    redirect_to @post
+  end
+
+
+
+  private
+  def comment_params
+    params.require(:comment).permit(:commenter,:commenter_nationality, :comment_content)
+  end
```
amaseda, line 46: Not that you need to worry about it for this assignment, but if you find that you're defining a lot of attributes with the same prefix -- in this case, `commenter` -- it means you might benefit from a separate `Commenter` or `User` model.

app/models/post.rb
```diff
@@ -0,0 +1,5 @@
+class Post < ActiveRecord::Base
+  has_many :comments
+  has_many :tags
+  has_many :categories, through: :tags
+end
```
amaseda, line 5: :+1:

amaseda:
:+1: Good job with this. Let me know if you want to talk about how to continue implementing the bonuses, especially in how they might relate to Project 2.

PR #209, opened 2016-06-30
Joe Calvillo Scribble HW

joecal:
Comfort: 4
Complete: 5

I kind of procrastinated and did it last minute so it's not pretty, but it works.


amaseda:
:+1: Great job here. Important thing is that you built a RESTful Rails app! Let me know if you have any lingering questions but it seems like you've got this stuff down.

PR #213, opened 2016-06-30
Really wish I had another day to work on this...

BlvckYuppie:
something clicked around 3 thurday morning, if this assignment were due friday, i'd finish it =(


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,17 @@
+class PostsController < ApplicationController
+  def index
+    @posts = Post.all
+  end
+
+  def show
+    @posts= Post.find.(params(:id))
```
amaseda, line 7: This will error out. It should be `Post.find(params[:id])` (with no period after `find`).

Also, a more apt variable name would be `@post` since `.find` will only return a single post.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  def new
+    @posts = Post.new
+  end
+  def create
+    # @posts = Post.find(params(:id))
```
amaseda, line 14: You'll need to use `.create` in order to make a post.

This will require setting up a `post_params` strong params method, as we've done in class. Let me know if you need any help with this.

app/models/comment.rb
```diff
@@ -0,0 +1,4 @@
+class Comment < ActiveRecord::Base
+
+belongs_to :posts
```
amaseda, line 3: Should be `belongs_to :post`, where post is written singularly.

app/views/posts/new.html.erb
```diff
@@ -0,0 +1,11 @@
+<h2>New Post</h2>
+
+<%= form_for @posts do |f| %>
```
amaseda, line 3: Again, would be very careful with how you name variables. You are creating a single new posts, not multiple ones (as `@posts` would imply).

db/migrate/20160628233646_create_posts.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePosts < ActiveRecord::Migration
+  def change
+    create_table :posts do |t|
+      t.string :title
+      t.text   :body
+      t.integer :post_id
```
amaseda, line 6: Rails will automatically generate a primary key. You don't need to do that.

You do, however, need to create a column for any foreign keys. That's not necessary for posts though...

db/migrate/20160628233655_create_comments.rb
```diff
@@ -0,0 +1,9 @@
+class CreateComments < ActiveRecord::Migration
+  def change
+    create_table :comments do |t|
+      t.text     :body
+      t.integer  :comment_id
```
amaseda, line 5: Same issue as with your posts migration -- Rails will automatically assign an id to a comment.

Since comments belong to posts, however, you do need to create a column for `post_id` so that we can attach a comment to a particular post.

db/schema.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  create_table "posts", force: :cascade do |t|
+    t.string   "title"
+    t.text     "body"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+  end
+
+end
```
amaseda, line 33: Your schema looks different than what I guess they would based on your migrations.

amaseda:
@BlvckYuppie We just had our 1:1 but, just to reiterate, I think you would benefit greatly from going through the [Getting Started with Rails tutorial](http://guides.rubyonrails.org/getting_started.html).

Also take a look at the comments I left here. There are several bugs/syntax errors that are causing errors.

PR #216, opened 2016-06-30
Kevin Golden Scribble

kwg58:
Completeness 2/5
Comfort 2/5

I got wrapped around the axle trying to put user functionality in.  I backed that out and got some progress done, but I'm just not as comfortable with Rails as I should be.


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    @comment = Comment.find(params[:id])
+  end
+
+  # update
+  def update
+    # redirect_to root_path unless @current_user
+    @comment = Comment.find(params[:id])
+    @comment.update(comment_params)
+
+    redirect_to "/comments/#{@comment.id}"
```
amaseda, line 41: How could you use a path helper here?

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  # new
+  def new
+    @post = Post.new
+  end
+
+  # create
+  def create
+    # redirect_to root_path unless @current_user
+    @post = Post.create!(post_params)
+    redirect_to "/posts/#{@post.id}"
```
amaseda, line 15: When dealing with RESTful actions, using path helpers in place of hard-coded paths makes your code a lot more readable.

app/models/comment.rb
```diff
@@ -0,0 +1,12 @@
+class Comment < ActiveRecord::Base
+  # belongs_to :user, dependent: :destroy
+  belongs_to :post, dependent: :destroy
+  validates :post, presence: true
```
amaseda, line 4: :+1:

amaseda:
Based on our conversations since this assignment, it sounds like you're more comfortable with setting up CRUD functionality. That pesky user auth got in the way here. Let me know if you have any lingering questions.

Also let me know if you want to talk more about user auth. I'm a fan of Devise. https://github.com/ga-wdi-lessons/devise

PR #217, opened 2016-07-01
Chris branch . 

gardner589:
comfort level - 9
completeness - 7


amaseda:
:+1: Good job here. Let me know if you have any particular questions about the assignment. If you get the chance try to implement some of those bonuses. That many-to-many one maybe?

PR #218, opened 2016-10-12
Domonique Griffin Scribble HW

griffind2009:
comfort: 4
completeness: 5


app/views/comments/index.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>List of Comments</h1>
+<% @comments.each do |comment| %>
+<%= comment.text %>
+<%= link_to 'Comments', comments_path %>
+<%= link_to 'Make a Comment', new_comment_path %>
+<%= link_to 'Edit your Comment', edit_comment_path(comment) %>
+<%= link_to 'Delete Comment', comment_path(comment), method: :delete %>
```
amaseda, line 7: It looks like comments are a nested resource in `routes.rb`. Because of this, you will need to use the proper path helpers, which will include `post` somewhere in their name. Take a look at `$ rails routes`...

config/routes.rb
```diff
@@ -0,0 +1,6 @@
+Rails.application.routes.draw do
+  root to: 'posts#index'
+  resources :posts do
+    resources :comments
```
amaseda, line 4: Make sure to use the correct nested path helpers for Comment elsewhere in your app.

amaseda (commented):
:+1:

PR #220, opened 2016-10-12
Oki Solution

nickoki:
comfort_level: 5
completeness: 5

Borrowed Views from inclass Scribble to focus more on the routes and controllers where I needed more practice


nolds9:
👍

PR #222, opened 2016-10-13
Scribble HW 14 (Michael L)

MichaelZLai:
Comfort: 4.5
Completeness: 5


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,42 @@
+class PostsController < ApplicationController
+  def index
+    @posts = Post.all
+  end
+
+  def show
+    @post = Post.find(params[:id])
+    @post_id = @post.id
```
nolds9, line 8: don't think you need to have a separate instance variable for this...can just reference `@post.id`...

nolds9:
👍

PR #236, opened 2016-10-13
Barrett Cope Submission

bcope:
Completeness: 3
Comfort level: 5


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    redirect_to comments_path
+  end
+
+  # strong params
+  private
+  def comment_params
+    params.require(:comment).permit(:text)
+  end
+end
```
amaseda, line 43: You've been able to practice this since completing this assignment but, nevertheless, how would you go about associating a comment with a particular post?

app/views/comments/new.html.erb
```diff
@@ -0,0 +1,8 @@
+<h1>New Comment</h1>
+<%= form_for :Comment, url: new_comment_path do |f| %>
```
amaseda, line 2: Is there a reason `:Comment` is capitalized here?

amaseda (commented):
:+1:

PR #238, opened 2016-10-13
BK Scribble

bkeating55:
complete -2
comfort-3


app/models/post.rb
```diff
@@ -0,0 +1,4 @@
+class Post < ActiveRecord::Base
+
+  has_many :comments
```
nolds9, line 3: no `Comment` model?

PR #239, opened 2016-10-13
scribble (Chris Jaffier)

chrisjaff:
comfort - 3
completeness - 2

I added what areas content in the areas I had the most knowledge on, but had trouble creating models,tables, and migrations. Please let me know what comment on what I have created.


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Comments < ApplicationController
+
+def index
+  @commennts = Comments.all
```
amaseda, line 4: Couple things...
- It should be `Comment`, not `Comments`. Very important you use singular notation when referencing the class/model name.
- You misspelled `@comments`. That could cause problems elsewhere if you're not consistent with naming.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+end
+
+def show
+  @commennts = Comments.find(:id)
```
amaseda, line 8: `@comments` again misspelled.

app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,9 @@
+class PostController < ApplicationController
+
+def index
+  @posts = Post.all
+end
+
+def show
+  @posts = Post.all
```
amaseda, line 8: The purpose of the `show` page is to display information about a single post. Here you're retrieving **all** of the posts in your database.

app/models/post.rb
```diff
@@ -0,0 +1,3 @@
+class Post < ActiveRecord::Base
+  has_many :comment
```
amaseda, line 2: Anything following `has_many` should be pluralized, like so...

``` rb
has_many :comments
```

app/views/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>Posts</h1>
+<%= @post.each do |post| %>
+<div class="post">
+  <h2><a href="post_path"</a>@post</h2>
```
amaseda, line 4: Your link isn't written out properly. Need to put some content between the `<a></a>` tags. Also need to close the opening tag.

config/routes.rb
```diff
@@ -0,0 +1,3 @@
+Rails.application.routes.draw do
+  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+end
```
amaseda, line 3: Don't see any routes in here. You want be able to use any of the path helpers (e.g., `posts_path`) referenced in your application without them.

Quick fix to start you off...

``` rb
resources :posts
```

amaseda (commented):
Good start but there are a number of things you need to fix to get this working properly. Take a look at my comments and please let me know if you have any questions / comments. Happy to talk about this!

PR #245, opened 2017-02-21
Lisa Burr

laburr27:
Confidence: 3.75
Completeness: DONE! 100% ! COMPLETE!

I feel so accomplished getting this done before it was officially due!!!!  It's ugly and nothing fancy but it does what it is supposed to do.  I feel much more confident in my rails abilities.  I must admit I used the various scribble gits to assist me in cleaning up my code and with troubleshooting.  But I am getting more proficient at recalling the correct information all the time.  Knowing what file does what and where and how it ties into functionality has finally clicked as have much of the ActiveRecord information.


superbuggy:
I'm glad you're feeling more comfortable, especially with nested resources in the mix! 👍

PR #249, opened 2017-02-23
Nat Carter | scribble

ghost:
Comfort: 4.4 / 5
Completeness: 5 / 5

Can you recommend any good sources on how to DRY up seeds?


amaseda:
Seed files tend to be long if you're coming up with all the data yourself. But gems like `Faker` can do all that work for you. With `Faker`, your seed file can just consist of a short enumerable(s).

https://github.com/stympy/faker

app/models/category.rb
```diff
@@ -0,0 +1,4 @@
+class Category < ApplicationRecord
+  has_many :tags
+  has_many :posts, through: :tags, dependent: :destroy
```
amaseda, line 3: :+1:

amaseda:
:+1:

PR #251, opened 2017-02-23
Scribble ~CR6 ~ Habtamu Abere

HabRonan:
Comfort- 3.9
completeness - 4.9
-didn't do the bonus


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,51 @@
+class CommentsController < ApplicationController
```
amaseda, line 1: Remember to be consistent with indentation!

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+  #index
+  def index
+  @post = Post.find(params[:post_id])
+  @comment = Comment.all
```
amaseda, line 5: This will give you all the comments. Seems like you only want all the comments that belong to a particular post though.

```rb
@comment = @post.comments.all
```

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,51 @@
...
+       redirect_to post_path(@post)
+     end
+
+
+def destroy
+  @post = Post.find(params[:post_id])
+  @comment = Comment.find(params[:id])
+  @comment.destroy
+
+redirect_to "/posts/#{@post.id}"
```
amaseda, line 41: Recommend using path helpers when possible.

db/migrate/20170222163607_create_posts.rb
```diff
@@ -0,0 +1,12 @@
+class CreatePosts < ActiveRecord::Migration[5.0]
+  def change
+    create_table :posts do |t|
+      t.string :title
+      t.string :author
+      t.string :content
+      t.integer :likes
+
+      t.timestamps :updated_at
```
amaseda, line 9: You can just say `t.timestamps`

amaseda:
:+1:

db/migrate/20170222163607_create_posts.rb
HabRonan: got it,thanks!

app/controllers/comments_controller.rb
HabRonan: thanks

app/controllers/comments_controller.rb
HabRonan: Thanks,
```
  def index
 @post = Post.find(params[:post_id])
  @comment = Comment.all
end
```
is just there if you wanna see all the comments on all the posts together, the show path is actually the one which will take you to that posts comment. But ya now that I think about it I didn't need to add that

app/controllers/comments_controller.rb
HabRonan: I definitely need to work on that.

app/controllers/comments_controller.rb
amaseda: Ha and now that I think about it, you're not using this route/controller action anyways, so doesn't make much of a difference either way!

PR #253, opened 2017-02-23
Aaron Andrews

andrews94:
completeness: 3

comfort: 3

Hard exercise and I couldn't quite finish.  Hopefully I will understand the remaining material soon.


app/controllers/show_controller.rb
```diff
@@ -0,0 +1,5 @@
+class ShowController < ApplicationController
```
superbuggy, line 1: this should rather be a method in the posts controller, and not it's own controller.

PR #254, opened 2017-02-23
Scribble — Rodney Curl

rodneycurl:
comfort 3
completeness 4

looking to seek more clarity on migrations those are still not my strong suit. overall this was a great exercise in helping me learn a bit about how to structure my files. now if only i can get everything else.


nayana487:
👍

nayana487:
migrations are what you use to generate your schema.

PR #257, opened 2017-02-23
Savraj Singh

SavrajSingh:


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,50 @@
+class CommentsController < ApplicationController
+
+  def index
+    @comments = Comment.all
```
amaseda, line 4: Should you be getting all the comments? Or all of the comments that belong to a particular post?

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  def new
+    @post = Post.find(params[:id])
```
amaseda, line 8: You'll need to use `:post_id` instead of `:id` to access the correct post. `:id` now corresponds with the comment's primary key -- not the post.

app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  def edit
+    @comment = Comment.find(params[:id])
+  end
+
+  # update
+  def update
+    @comment = Comment.find(params[:id])
+    @comment.update(comment_params)
+
+    redirect_to post_comment_path(@comment)
```
amaseda, line 34: You also need to pass in a post to this path helper. So it would look like...

```rb
post_comment_path(@post, @comment)
```

amaseda:
:+1:

PR #259, opened 2017-02-23
Scribbles HW - Elena Santana

fanelenan7:
Completeness - 5
Comfort level - 4


app/controllers/comments_controller.rb
```diff
@@ -0,0 +1,46 @@
+class CommentsController < ApplicationController
+
+def index
+  @post_comments = Comments.all
```
amaseda, line 4: The class should be `Comment` -- not `Comments`

amaseda:
:+1:

PR #264, opened 2017-02-24
Mitchell Carroll- Scribble

mitchellcarroll:
Comfort level: 2
Completeness: 3

Really struggled with this. I tried to code along with a previous tunr exercise, got lost and couldn't figure out how to get the app working properly. I got to the part where I was trying to add routes but only ended up adding one (which caused an error). I will continue to work on this over the weekend to get the practice.


AndyWhitley:
👍

PR #267, opened 2017-02-24
jays scribble-y

xgironx:
comf:  2/5 on create/update of nested resource items
comp:3/5, i really only got index & show working 
whew!  could use more work on nested resources, felt really good after officehours time on hogwarts only to hit the brick wall on scribble nesteds


superbuggy:
The path helper method are different in name, but also in what they take in terms of arguments. you have to provide an instance of a 'parent' object, and in addition, sometimes the 'child' object as well.

PR #268, opened 2017-02-24
Ben Shapiro - Scribble HW

pretorben:
Completeness: 5
Comfort Level: 4

The functionality is there, although I'm not sure whether or not the Scribble logo was supposed to be on the page and the style is bland.  I briefly attempted Devise but had issues (likely from rushing and just making syntax errors) so I took it back out and focused on just completing the assignment minus the bonuses.


superbuggy:
👍

PR #269, opened 2017-02-24
Scribble HW (Justin Shin)

Justin-Shin:
Completeness: 3
Comfort Level: 5

Completeness is at 3 because I realized I forgot to do the comment portion.  Oops.

DIdn't make it pretty, saving that juice for GoT and the project.  I'm extremely comfortable with rails now, the challenge will be making it do more complex tasks for the project.


superbuggy:
👍

PR #272, opened 2017-02-24
Jared Edwards HW Scribble

jarededwards:
Completeness: 3 
Comfort Level : 4

when I went to push my repo this morning I realized I cloned from the instructions "$ git clone git@github.com:ga-wdi-exercises/scribble.git" and was unable to push my homework do to permission issues. I should have made an initial push earlier on to avoid this but this is what I am stuck with.

I am going to try and get as much done in the next half hour ish to have something to show.


app/views/comments/destroy.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1>Comments#destroy</h1>
+<p>Find me in app/views/comments/destroy.html.erb</p>
```
nayana487, line 2: what is this?

nayana487:
hey, schedule a one-on-one next week if ya wanna go through this. doesn't look like a lot of this would work if you don't have a controller with actions

PR #274, opened 2017-02-24
Clements - Scribble

sdkclements:
Had some weird issues started overa bunch


config/routes.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  get 'posts/show'
+
+  get 'posts/create'
+
+  get 'posts/update'
+
+  get 'posts/delete'
+
+	resources :comments, :posts
```
amaseda, line 23: Don't need the individual routes above if you're using `resources`

amaseda (dismissed):

amaseda:
Let me know if you have questions about this one. Noticed some files (like your controllers) are empty, which might be due to starting over.

PR #281, opened 2017-04-22
Scribble - Ick Hyun Kwon

kwonih1020:
comfort - 4
completeness - 5 (except bonus)

Delete and redo for 3times because keep getting errors on rails db:migrate and seed part.
It felt pretty easy on last trial, need more understanding on db:migrate and db:seed part.


juancgarcia (commented):
👍 Good job

It's pretty common to have trouble with your database when first working with migrations and seeds. Practice helps you get a better understanding.

PR #283, opened 2017-04-24
scribble HW - Stormy Ramsey

StormyRamsey:


juancgarcia (commented):
👍 Good work

I'm not quite sure why you changed from `Post` to `Article` midway through the assignment. It still works but strictly speaking, the assignment did call for the model to be `Post`

In the future, when creating a rails app with `rails new` , make a commit (`git add .` and `git commit -m "message"`) before you start working on your changes. That makes it easier to discriminate your work from the generated rails files

PR #284, opened 2017-04-24
Hali Bakarr - Scribble

utamilah:
Comfort 4.5
Completeness 4.5
––––
I had a lot of issues with redirecting after editing comments but I believe I fixed it. Also, my 'if' statement for the comments in the posts "show" file doesn't seem to be working. It's meant to display "Post a comment!" when there aren't any comments on the page. I wonder why.


AndyWhitley:
It's because your testing the association against nil. If there are no comments, `@post.comments` will still evaluate to an empty array `[]` (which is not `nil`). You could test the length instead:
```erb
<% if @post.comments.length == 0 %>
```

👍

PR #286, opened 2017-04-24
First commit

pchristianbrown:


jsm13:
Looks like this started out on the pre-devise branch rather than the master branch and something funny happened with moving a lot of files around. We're you able to get this running or test at all?

PR #287, opened 2017-04-24
Scribble Jeong Hoon Koo

jkoo87:
Comfort: 3.5
Completeness: 5


juancgarcia (commented):
👍 Nice work.

In the future, when creating a rails app with `rails new` , make an inital commit (ex: `git add .` and `git commit -m "new rails app"`) before you start working on your changes. That makes it easier to discriminate your work from the generated base rails files

PR #290, opened 2017-04-25
scribble

zonekid:
difficulty-3.5
comfort- 4

I started the assignment last week which is why its just the basic formula.  But I stuck to it because it gave me the opportunity to kind of rewind and implement the basics which was a challenge, instead of picking up from where we were in class and implementing those tools.


juancgarcia (commented):
👍 

In the future, when creating a rails app with `rails new` , make an inital commit (ex: `git add .` and `git commit -m "new rails app"`) before you start working on your changes. That makes it easier to tell your work from the generated rails files

PR #291, opened 2017-04-25
Scribble HW 04/24

edawson1980:


edawson1980:
Completeness: 4
Comfortability: 4

I was able to get everything to work the way I wanted it to, with the exception of my comments showing up and formatting the logo.  Overall, I feel pretty comfortable.  I'm having some difficulty keeping track of my paths and routes, but I think that comes from having trouble keeping track of what needs to be pluralized and what does not.  I definitely understand the concepts - I'm trying to nail down all the fine details now.

jsm13:
Looks great! Using `rails routes` can be super helpful in both checking what helper methods to use (prefix will be in the far left column). Looking at this regularly should help develop a feel for when names should be singular or plural

PR #292, opened 2017-04-25
WPurdum Scribbler HW - with bonus

wjpurdum:
Completion 5/5 - would have done this through user authentication if I had started it later!
Comfort level 4.5/5 - the workflow (controller, view, route) and logic makes sense. Only tricky part for me is when to be sure I must pass in arguments (one or both, and if they need to have an @ preceding it).
Attempted bonus.


wjpurdum:
I think I thwarted myself a bit as I added a "category" to my posts when I first seeded the database, not realizing that would be a component of the bonus material.

juancgarcia (commented):
👍 Good work.

About passing arguments, I presume you're referring to url/path helper methods. When you're using nested routes, that nested approach implies you're stepping through a list of objects while building out the path.

For example the path `/posts/3/comments/2` corresponds to the following rails route (found via the `rails routes` command or `http://localhost:3000/rails/info/routes` from your running rails server):
```ruby
# generated routes
post_comment_path   /posts/:post_id/comments/:id(.:format)   comments#show
```
The original path has two numeric variables (3, 2) matching the named params in the route listing (:post_id, :id). Those two id numbers come from two object parameters supplied to the `post_comment_path` helper method. Something like:
```ruby
post_comment_path(@comment.post, @comment)
```

If your path only calls for one `id` like `/posts/4`
```ruby
# generated routes
post_path   /posts/:id(.:format)   posts#show
```
then `post_path` only expects one object from which to extract the id:
```ruby
post_path(@post)
```
 In short, it's understanding what path you want to create, and then supplying objects with the ids needed to generate that path.


Also, in the future, when creating a new app via `rails new` or some of the other tools we'll use soon, make an inital commit (ex: `git add .` and `git commit -m "first commit"`) before you start working on your changes. That makes it easier to discriminate your work from the generated base rails files.

PR #296, opened 2017-04-25
scribble changes

lgmusic24:
Plenty of trouble with this. I had trouble loading rails new . -d postgreslq and connecting my new files. I will probably work on this exercise this weekend.


jsm13:
Have you done more work on this that could be pushed up? There are no migrations in this PR but the controllers are all built out. Be very careful about writing code before things are working. Without the database set up and the models working there is no way to run the app between changes. Later when the DB is set up, mistakes in the controller will be hard to track down as the code was added in large untested portions rather than small tested steps

PR #299, opened 2017-06-07
Scott's scribble

scottpiercew:
This is the error I have not yet got past.
ActiveRecord::RecordNotFound in CommentsController#edit
Couldn't find Comment with 'id'=7

```  def edit
    @post = Post.find(params[:post_id])
    @comment = Comment.find(params[:id])
  end

  def update```

### CODE REVIEW

File: app/views/posts/show.html.erb [hunk 1]
```diff
@@ -0,0 +1,16 @@
...
+<p><%= @post.body %></p>
+<p><%= link_to "Edit Post", edit_post_path(@post) %></p>
+<p><%= link_to "Delete Post", post_path(@post), method: :delete %></p>
+
+<p><%= link_to "Add new comment", new_post_comment_path(@post) %></p>
+
+
+<% @post.comments.each do |comment| %>
+<p><%= comment.body %></p>
+<p><%= link_to "Edit comment", edit_post_comment_path(@post) %></p>
```
superbuggy, line 14: The error you described is coming from Line 14. Rails is reading the post's id as the comments id (imo this behavior is dumb, it should throw an error* -- scroll to the end for my explanation of this thought). You need to pass in one more argument, the comment, so that the helper method `edit_post_comment_path` 'knows' what comment to edit. 

Fix: `edit_post_comment_path(@post, comment)`

*Since that path helper method takes two arguments, I think it should throw an error when it gets only 1.

superbuggy:
👍

PR #303, opened 2017-06-07
Scribble

daniellefricke:
Comfort 3 - I get most of the concepts, just need more practice to feel more comfortable. 
Complete 4

There is an error in the posts show page that I couldn't get past. I'm sure there is more debugging after that.


app/controllers/posts_controller.rb
```diff
@@ -0,0 +1,41 @@
+class PostsController < ApplicationController
+
+    def index
+        @post = Post.all
```
amaseda, line 4: Would recommend calling this `@posts` since it will contain multiple posts.

app/models/comment.rb
```diff
@@ -0,0 +1,4 @@
+class Comment < ApplicationRecord
+  belongs_to :post
+  validates :name, :text, {presence: true}
```
amaseda, line 3: :+1:

app/models/post.rb
```diff
@@ -0,0 +1,4 @@
+class Post < ApplicationRecord
+  has_many :comments, dependent: :destroy
+  validates :title, :author, :text, {presence: true}
```
amaseda, line 3: You shouldn't need to use curly brackets here. Rails will assume you are writing out a hash if the last argument(s) of a method are key-value pairs.

app/views/posts/index.html.erb
```diff
@@ -0,0 +1,11 @@
+<h1>Scribble</h1>
+<h4><a href="/posts/new">Add a Scribble</a></h4>
+
+
+<% @post.each do |post| %>
+  <li>
+    <a href="/posts/<%=post.id%>">
```
amaseda, line 7: Encourage you to use `link_to` and `path helpers` whenever possible.

db/migrate/20170605003515_create_posts.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePosts < ActiveRecord::Migration[5.1]
+  def change
+    create_table :posts do |t|
+      t.string :title
+      t.string :text
+      t.string :author
+      t.string :date
```
amaseda, line 7: Consider using the `.datetime` datatype so you have access to Ruby's datetime methods.

PR #304, opened 2017-06-07
Scribble.

BaoPham92:
I removed my migration, seed, and schema. I was in the middle of redoing, I believe I did the setup wrong.


superbuggy:
What led to remove your migrations and schema? In a real bind, ***if you don't need your data***, you can always drop your db data `rails db:drop` and just edit your migrations, then re-run. It looks like a lot of this is functional.

PR #305, opened 2017-06-07
AntwanaW

Antwana:


app/models/post.rb
```diff
@@ -0,0 +1,3 @@
+class Post < ApplicationRecord
+  has_manu :comments
```
superbuggy, line 2: `has_many`

superbuggy:
it looks like you hit a snag related to Active Record or migrations.

PR #306, opened 2017-06-07
scribble

Fitssum:


app/views/comments/index.html.erb
```diff
@@ -0,0 +1 @@
+<h2>Hello there!</h2>
```
superbuggy, line 1: Probably don't need an index view for comments, just because you wouldn't want to view them all together as a list, but without their posts. If you were viewing them along with the posts they belong to, it would essentially be the same as a posts index view.

app/models/post.rb
```diff
@@ -0,0 +1,4 @@
+class Post < ApplicationRecord
+  has_many :comments
+  validates :post_name, :topic, :post_url, {presence: true}
```
superbuggy, line 3: validations, nice!

superbuggy:
👍

PR #307, opened 2017-06-07
scribble

tedlayea:


config/routes.rb
```diff
@@ -0,0 +1,5 @@
+Rails.application.routes.draw do
+  # For details on the DSL available within this file, see http://guides.rubyonrails.org/routing.html
+  resources :post
+  resources :commentss
```
amaseda, line 4: Looks like you have an extra "s" at the end of `:comments`

db/migrate/20170607001242_create_posts.rb
```diff
@@ -0,0 +1,10 @@
+class CreatePosts < ActiveRecord::Migration[5.1]
+  def change
+    create_table :posts do |t|
+      t.string :title
+      t.string :mainpost
+      t.datetime "created_at"
```
amaseda, line 6: Why did you wrap `created_at` and `updated_at` in quotes?

db/migrate/20170607001253_create_comments.rb
```diff
@@ -0,0 +1,10 @@
+class CreateComments < ActiveRecord::Migration[5.1]
+  def change
+    create_table :comments do |t|
+      t.string :comment
```
amaseda, line 4: I would recommend naming this attribute something other than the name of the model itself (e.g., body, text).

PR #315, opened 2017-08-03
scribblez

jacobg1:
5
5


juancgarcia:
👍 Good work!

I suppose in this case where your comfort and completeness are both 5, the labels don't really matter, but try to keep them in there.

